PR #6, opened 2025-10-22
feat: introduces openai_responses chat provider and supports think part UI

Epsirom:
Blocked by: https://github.com/MoonshotAI/kosong/pull/1


Epsirom:
Pls @stdrc have a look.

Perhaps the thinking UI can be better.

PR #45, opened 2025-10-24
docs: better change log version

yihong0618:
as diff....


stdrc:
What's the rationale behind this🧐

yihong0618:
> What's the rationale behind this🧐

because we have a release notes on command

and the last one is the biggest its confusing


thinking about the version make sense here, closing this


image

PR #54, opened 2025-10-24
fix(glob): use f-string for no matches message

QIN2DIM:
The message when no matches are found was not using an f-string, which could lead to inconsistent formatting. This change ensures the message uses an f-string for consistent and correct string interpolation, matching the style of the success message.


.gitignore
```diff
@@ -19,3 +19,4 @@ uv.toml
 # Build dependencies
 src/kimi_cli/deps/bin
 src/kimi_cli/deps/tmp
+.idea/*
```
stdrc: Please move to `# Project files` section

stdrc:
Actually we'd better have a unit test for this

QIN2DIM:
1

stdrc (approved):
LGTM

PR #70, opened 2025-10-24
Fix bash decode

YangLi-leo:


stdrc:
This is already merged with https://github.com/MoonshotAI/kimi-cli/commit/f78f83fd39310795b381ecad7f46d5e28f4395e7

YangLi-leo:
i just saw some test failed, so i change the format of "replace" in order to fit with the code style

stdrc:
> so i change the format of "replace" in order to fit with the code style

I already formatted the code when I was trying to merge it directly on my local machine

PR #72, opened 2025-10-24
feat: add Windows support for directory listing

YangLi-leo:
Adding Windows support for the directory listing feature by using the `dir` command on Windows while maintaining `ls -la` on Unix-like systems (macOS/Linux).
I also update the system prompt in order to make the model not be confused by the output of the directory listing.


stdrc (approved):
LGTM

Issue #64, opened 2025-10-24
add Openrouter support


hadifarnoud:
### What feature would you like to see?

I cannot use openrouter API keys. as we are using them for everything else, it would make it super easy to use kimi cli for me

xingty:
kimi-cli supports APIs that are compatible with openai's. To use those APIs, you can edit your config file (located at ~/.kimi) as follows:

```json
{
  "default_model": "",
  "models": {
    "gpt4.5": {
      "provider": "openai",
      "model": "gpt-4.5",
      "max_context_size": 160000
    }
  },
  "providers": {
    "openai": {
      "type": "openai_legacy",
      "base_url": "base_url",
      "api_key": "your-secret-key"
    }
  },
  "loop_control": {
    "max_steps_per_run": 100,
    "max_retries_per_step": 3
  },
  "services": {}
}
```
and then run kimi with:
```shell
kimi -m gpt4.5
```

stdrc:
Closing as it's already answered.

hbprotoss:
> kimi-cli supports APIs that are compatible with openai's. To use those APIs, you can edit your config file (located at ~/.kimi) as follows:
> 
> {
>   "default_model": "",
>   "models": {
>     "gpt4.5": {
>       "provider": "openai",
>       "model": "gpt-4.5",
>       "max_context_size": 160000
>     }
>   },
>   "providers": {
>     "openai": {
>       "type": "openai_legacy",
>       "base_url": "base_url",
>       "api_key": "your-secret-key"
>     }
>   },
>   "loop_control": {
>     "max_steps_per_run": 100,
>     "max_retries_per_step": 3
>   },
>   "services": {}
> }
> 
> and then run kimi with:
> 
> kimi -m gpt4.5

is it replaced by `config.toml`?

PR #80, opened 2025-10-25
feat(provider): Add OpenRouter support  (#64)

TheNomadInOrbit:
Add OpenRouter provider with flexible setup options via environment variables and config file. Includes support for:

- Base URL configuration
- API key management
- Custom headers
- Model selection
- Context size

Makes it straightforward for users to connect OpenRouter models to the Kimi CLI with just a few config tweaks.


stdrc:
We currently don't have plan to build in such third-party gateway support, so closing this for now. You can simply use the `openai_legacy` provider type for OpenRouter API.

PR #82, opened 2025-10-25
fix: Handle CancelledError in bash and task tools

YangLi-leo:
## Summary
Fix resource leak caused by orphan processes when operations are cancelled.

## Changes
 - Add CancelledError handling in bash tool to clean up subprocesses properly
 - Add CancelledError handling in task tool for subagent cancellation
 - Return proper ToolError instead of propagating exception


YangLi-leo:
I think the reason i failed CI is the refactoring of Wire to WireUI, The core fix for CancelledError is correct. Happy to update for the new API if needed.

stdrc:
> I think the reason i failed CI is the refactoring of Wire to WireUI, The core fix for CancelledError is correct. Happy to update for the new API if needed.

Yes I'm refactoring the codebase to prepare for Kimi CLI SDK. I'll update you once I complete, you can resolve conflicts then.

stdrc:
BTW could you provide a reproducible case that this CancelledError is somehow visible to the user? IIUC cancellation of a tool call task will be [properly handled by `kosong`](https://github.com/MoonshotAI/kosong/blob/main/src/kosong/__init__.py#L41)

YangLi-leo:
you are right, the CancelledError can be solved by https://github.com/MoonshotAI/kosong/blob/main/src/kosong/__init__.py#L41, but this part can solve the FIXME comment about processing cleanup while letting kosong handle the error response:     except asyncio.CancelledError:
        # Handle user cancellation (Ctrl+C or ESC key)
        process.kill()
        with contextlib.suppress(Exception):
            # Wait for process to exit with protection, ignore cleanup failures
            await asyncio.shield(asyncio.wait_for(process.wait(), timeout=2.0))
        raise

stdrc:
> but this part can solve the FIXME comment about processing cleanup

Which FIXME comment?

YangLi-leo:
this one you can check it: [https://github.com/MoonshotAI/kimi-cli/blob/main/src/kimi_cli/tools/bash/__init__.py](url)  line 83

stdrc:
close as outdated

PR #113, opened 2025-10-28
feat(provider): Add ZenMux

henryhe1993:
[ZenMux](https://zenmux.ai/) is the world’s first enterprise-grade large model aggregation platform with an insurance payout mechanism. The platform provides one-stop access to the latest models across providers. 

Add ZenMux as a provider to allow users to access ZenMux OpenAI-Compatible models(60+) in the cli. Support configurations of:
- Base URL
- API key
- Model
- Context size

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,29 @@ For example, to use Kimi CLI with [Zed](https://zed.dev/), add the following con
 
 Then you can create Kimi CLI threads in Zed's agent panel.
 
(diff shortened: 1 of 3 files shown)


stdrc:
`openai_legacy` provider type can be used for any openai-compatible api platforms.

stdrc:
We currently don't have plan to build in such third-party gateway support, so closing this for now.

PR #137, opened 2025-10-31
docs: add uvx quick start doc

yihong0618:
as diff 

for qucik doc


xxchan:
I think we'd rather encourage users to install kimi-cli instead.

PR #156, opened 2025-11-04
  fix(makefile): use defined RUFF and PYRIGHT variables

winterfx:
Replace hardcoded "uv run ruff" and "uv run pyright" commands with the defined $(RUFF) and $(PYRIGHT) variables in format and check targets.


stdrc:
Actually we are just about to remove the defined variables... For better predictability

winterfx:
No problem, I’ll close this PR. Thanks!

PR #160, opened 2025-11-05
feat(glob): return ls-style work dir listing for '**' pattern

bigboss2063:
When using the Glob tool, if the pattern is **, give an `ls -a` style result for the workdir


src/kimi_cli/tools/file/glob.py
```diff
@@ -1,17 +1,54 @@
...
+
+def _lookup_group(gid: int) -> str:
+    """Resolve a gid to a user-friendly name."""
+    if grp is None:  # pragma: no cover - Windows fallback
+        return str(gid)
+
+    try:
+        return grp.getgrgid(gid).gr_name
+    except KeyError:  # pragma: no cover - gid without entry
+        return str(gid)
```
stdrc: looks unnecessary to me

src/kimi_cli/tools/file/glob.py
```diff
@@ -45,10 +82,9 @@ def __init__(self, builtin_args: BuiltinSystemPromptArgs, **kwargs: Any) -> None
     async def _validate_pattern(self, pattern: str) -> ToolError | None:
         """Validate that the pattern is safe to use."""
         if pattern.startswith("**"):
-            # TODO: give a `ls -la` result as the output
-            ls_result = await aiofiles.os.listdir(self._work_dir)
+            ls_result = await self._format_workdir_listing()
```
stdrc: I think we can just reuse `_list_work_dir` in `runtime.py` (move it to `utils/path.py` and name it `list_directory`.

src/kimi_cli/tools/file/glob.py
bigboss2063: fixed

src/kimi_cli/tools/file/glob.py
bigboss2063: fixed

bigboss2063:
@stdrc Hi, I fixed the code according to your suggestion

Issue #146, opened 2025-11-03
Adding a "paste collapse" feature


wilbeibi:
### What feature would you like to see?

When pasting large text (logs, code, documents), the terminal gets flooded. It would be helpful to show a collapsed summary like  "pasted: 500 words" instead, similar to Claude CLI and Codex.

Happy to submit a PR if this fits the project's direction.

yihong0618:
yes this is indeed we need. I love codex with `paste collapse` feature

PR #162, opened 2025-11-06
[feat] Add support for paste collapse  || [feat] Add support for paste collapse

wilbeibi:
This PR collapse large text paste into a placeholder similar to what image is doing. It also makes backspace/delete behavior nicely w/ these placeholders, when backspace, it delete the whole placeholder instead each char.

fixes https://github.com/MoonshotAI/kimi-cli/issues/146 

---
This PR collapse large text paste into a placeholder similar to what image is doing. It also makes backspace/delete behavior nicely w/ these placeholders, when backspace, it delete the whole placeholder instead each char.

fixes https://github.com/MoonshotAI/kimi-cli/issues/146

### DIFF
--- a/src/kimi_cli/ui/shell/prompt.py
+++ b/src/kimi_cli/ui/shell/prompt.py
@@ -394,6 +394,8 @@ def toast(message: str, duration: float = 5.0) -> None:
     _toast_queue.put_nowait((message, duration))
 
 
+# Matches both image and text placeholders:
+# [image:id,640x480] or [text:id,60 lines]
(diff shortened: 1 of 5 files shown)


stdrc:
I think in Codex CLI when the folded message is "sent", the original whole input will be printed. Is it true in this PR? If that is the case, it seems we don't need special logic in `message_stringify` when replaying session right?

stdrc:
I did some refactoring last week so there're some conflicts to be resolved.

stdrc:
close as outdated

Issue #27, opened 2025-10-23
feat: tool call result visualization


stdrc:

xxchan:
https://github.com/MoonshotAI/kimi-cli/pull/638

PR #165, opened 2025-11-06
feat: file changes preview

xingty:
This PR introduces a preview feature to display file changes in the terminal. (maybe relates to #27)

## Changes
- File change details can now be displayed directly in the terminal
- Added new components: `PreviewChange`, `Preview`, and a queue for UILoop handling
- Implementation follows a similar pattern to the existing request approval feature

image

## Discussion
While the current implementation works, I've realized that using `PreviewChange` as a payload of `ApprovalRequest` might lead to a cleaner architecture with fewer changes. This alternative approach could simplify the code structure.

Which approach do you think is better? I can implement the alternative in a new commit if the team prefers that direction.


xxchan:
https://github.com/MoonshotAI/kimi-cli/pull/638

PR #169, opened 2025-11-06
refactor: enable Pyright strict type checking for kimi_cli/ui/shell

2725244134:
#168
## Summary

Enable Pyright strict type checking for the shell UI and soul modules by:
1. Adding explicit type annotations (e.g., `PromptSession[str]`, `asyncio.Task[None]`)
2. Creating public interfaces for protected members to satisfy `reportPrivateUsage` checks

### Commit 1: Type Annotations
- Enable strict type checking for `src/kimi_cli/ui/shell/` directory in pyrightconfig.json
- Add explicit type parameters to generic types
- Add `__all__` export list in metacmd.py to resolve import warnings

### Commit 2: Public Interfaces
- Add public properties to `KimiSoul`: `agent`, `runtime` 
- Add public property to `Context`: `file_backend`
- Add public method to `ShellApp`: `run_soul_command()`
- Update all usages to use public interfaces instead of protected members

### DIFF
--- a/src/kimi_cli/soul/kimisoul.py
+++ b/src/kimi_cli/soul/kimisoul.py
@@ -98,7 +98,7 @@ class KimiSoul(Soul):
     @property
     def agent(self) -> Agent:
         return self._agent
-    
+
     @property
     def runtime(self) -> Runtime:
         return self._runtime
--- a/src/kimi_cli/ui/shell/__init__.py
+++ b/src/kimi_cli/ui/shell/__init__.py
@@ -155,7 +155,6 @@ class ShellApp:
             console.print(f"[red]Unknown error: {e}[/red]")
             raise  # re-raise unknown error
 
-
     async def run_soul_command(self, user_input: str | list[ContentPart]) -> bool:
         """public method to run soul command."""
         return await self._run_soul_command(user_input)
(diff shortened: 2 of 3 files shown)


src/kimi_cli/ui/shell/metacmd.py
```diff
@@ -245,7 +245,7 @@ async def compact(app: "ShellApp", args: list[str]):
     """Compact the context"""
     assert isinstance(app.soul, KimiSoul)
 
-    if app.soul._context.n_checkpoints == 0:
+    if app.soul.context.n_checkpoints == 0:
         console.print("[yellow]Context is empty.[/yellow]")
         return
```
stdrc, line 250: I'm planning to move all kimi-soul-only meta commands into `kimisoul`, which means to introduce soul-level meta commands. Will delay this PR.

2725244134: get it

stdrc: Well, it seems I don't have bandwidth to design soul-level meta commands in recent week. What about let's enable strict mode for all `shell` files except `metacmd.py` in this PR?

2725244134: OK

Issue #174, opened 2025-11-06
Read before edit


alvis-HaoH:
### What feature would you like to see?

Ensure the model reads the file at least once before editing it.

Eric-Guo:
So the `kimi-cli` should make sure any file the LLM try to using tools in `StrReplaceFile` is being at least using tools `ReadFile` first? But what if the LLM is already using `cat` or `head` command in `Bash`? IMHO, it should also allow, so that's the other agent beehives in such case?

PR #175, opened 2025-11-06
Refactor: Enforce Read prerequisite for patch/replace tools

alvis-HaoH:
## Related Issue

Resolve #(174)

## Description
  - Introduce a session-scoped `FileOpsWindow` that tracks which files have been read via `ReadFile`.
  - Inject the shared window through the agent runtime so every file tool instance shares the same state.
  - Update `ReadFile` to mark files as read after a successful call, and make `PatchFile`/`StrReplaceFile` refuse edits unless
 the file was read earlier in the session.
  - Refine user-facing docs to call out the new “read before edit” requirement and expand pytest coverage for both happy-path and failure scenarios.

## Testing
  - uv run pytest tests/test_patch_file.py tests/test_str_replace_file.py -vv


xxchan:
Thanks. But we don't see a need for this now. We may discuss in #174

PR #204, opened 2025-11-09
feat: prefer basename match in FileMentionCompleter

xxchan:
Before
image
After
image

### DIFF
--- a/tests/test_file_completer.py
+++ b/tests/test_file_completer.py
@@ -109,7 +109,9 @@ def test_basename_prefix_is_ranked_first(tmp_path: Path):
     texts = _completion_texts(completer, "@fetch")
 
     # Snapshot the full candidate list to keep order/content deterministic
-    assert texts == snapshot([
-        "src/kimi_cli/tools/web/fetch.py",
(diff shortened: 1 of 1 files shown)


src/kimi_cli/ui/shell/prompt.py
```diff
@@ -314,7 +314,26 @@ def get_completions(self, document, complete_event):
...
-            yield from self._fuzzy.get_completions(mention_doc, complete_event)
+            # First, ask the fuzzy completer for candidates.
+            candidates = list(self._fuzzy.get_completions(mention_doc, complete_event))
+
+            # re-rank: prefer basename matches
+            frag_lower = fragment.lower()
+
+            def _rank(c: Completion) -> tuple:
+                path = c.text
+                base = path.rstrip("/").split("/")[-1].lower()
```
yihong0618, line 325: this seems not right on windows

xxchan: It seems we used `as_posix()` when building the completion candidates

yihong0618: cool learned thanks

Issue #264, opened 2025-11-13
Kimi CLI crashes if LLM Provider returns None for the usage "input_other" tokens


ZakWork:
### What version of Kimi CLI is running?

kimi, version 0.53

### Which open platform/subscription were you using?

Chutes

### Which model were you using?

Kimi K2 Thinking

### What platform is your computer?

Windows

### What issue are you seeing?

Kimi CLI crashes if LLM Provider returns None for the usage "input_other" tokens 

site-packages\kosong\chat_provider\__init__.py:93 in input                                            │
│                                                                                                                                                             │
│    90 │   @property                                                                                                                                         │
│    91 │   def input(self) -> int:                                                                                                                           │
│    92 │   │   """Total input tokens, including cached and uncached tokens."""                                                                               │
│ ❱  93 │   │   return self.input_other + self.input_cache_read + self.input_cache_creation                                                                   │
│    94                                                                                                                                                       │
│    95                                                                                                                                                       │
│    96 type ThinkingEffort = Literal["off", "low", "medium", "high"]                                                                                         │
│                                                                                                                                                             │
│ ╭───────────────────────────────────────── locals ──────────────────────────────────────────╮                                                               │
│ │ self = TokenUsage(input_other=None, output=0, input_cache_read=0, input_cache_creation=0) │        

### What steps can reproduce the bug?

Use Chutes claude endpoint "https://claude.chutes.ai" and set the model config to have thinking on < "capabilities": ["thinking"] >. If you send any message you get the error above.

### What is the expected behavior?

Not crash and assume token usage as not reported/0

stdrc:
Please report to the Chutes vendor that their API is not OpenAI-compatible.

PR #265, opened 2025-11-13
fix: handle None input_other in token usage to stop cli from crashing

ZakWork:
Add defensive handling for cases where result.usage.input_other is None by setting it to 0. This prevents potential issues when provider returns token counts as None and ensures the app does not crash.

## Related Issue
#264 

Resolve #(issue_number)

## Description

Added a check if the input_other field is None it gets replaced by 0

### DIFF
--- a/src/kimi_cli/soul/kimisoul.py
+++ b/src/kimi_cli/soul/kimisoul.py
@@ -4,7 +4,7 @@ import asyncio
 from collections.abc import Sequence
 from functools import partial
 from typing import TYPE_CHECKING
-
+from dataclasses import replace
 import kosong
 import tenacity
 from kosong import StepResult
@@ -233,6 +233,12 @@ class KimiSoul(Soul):
         logger.debug("Got step result: {result}", result=result)
         if result.usage is not None:
             # mark the token count for the context before the step
+            if result.usage.input_other is None:
+                logger.warning("input_other is None in token usage, setting to 0")
+                # Create new usage with input_other=0
+                new_usage = replace(result.usage, input_other=0)
+                # Create new StepResult with updated usage
+                result = replace(result, usage=new_usage)
             await self._context.update_token_count(result.usage.input)
             wire_send(StatusUpdate(status=self.status))
 


src/kimi_cli/soul/kimisoul.py
```diff
@@ -233,6 +233,12 @@ async def _kosong_step_with_retry() -> StepResult:
         logger.debug("Got step result: {result}", result=result)
         if result.usage is not None:
             # mark the token count for the context before the step
+            if result.usage.input_other is None:
+                logger.warning("input_other is None in token usage, setting to 0")
+                # Create new usage with input_other=0
+                new_usage = replace(result.usage, input_other=0)
+                # Create new StepResult with updated usage
+                result = replace(result, usage=new_usage)
             await self._context.update_token_count(result.usage.input)
```
yihong0618, line 242: result.usage.input or 0
is better?

ZakWork: It should be 0

https://github.com/MoonshotAI/kosong/blob/4c25822441fda91bea52dbd10f16365883996ea8/src/kosong/contrib/chat_provider/anthropic.py#L260
image

The actual issue is in the code above, there should be a "or 0" at the end of input_other=self._usage.input_tokens, I have made the pull request for that as well as otherwise when the sum is made in the code below it will raise the error 
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'

https://github.com/MoonshotAI/kosong/blob/4c25822441fda91bea52dbd10f16365883996ea8/src/kosong/chat_provider/__init__.py#L93
image


Once that is accepted/integrated in, this won't be needed here but in the meantime any provider not reporting back the input token with the Anthropic API will just crash the Kimi CLI.

yihong0618: got it, thanks for the info, you are right

stdrc:
`input_other` is defined as `int` in `kosong.chat_provider.TokenUsage`, I don't think there should be any chance that it is set to None if pyright check is enabled.

stdrc:
According to the OpenAI SDK definition of `CompletionUsage`, `completion_tokens` is also defined as `int`. Please report to the Chutes vendor that their API is not OpenAI-compatible.

ZakWork:
This is for the Anthropic API endpoint but your point still makes sense as the Anthropic API also requires and integer for input tokens. However, the Kimi CLI should not crash and exit because the usage input_token is missing and instead gracefully deal with the problem. Claude Code deals with this and the Chutes Anthropic API endpoint has no such issue there. 

Therefore, to make the Kimi CLI more robust either this change and/or the pull request in the Kosong library (https://github.com/MoonshotAI/kosong/pull/21) should be made 

image

stdrc:
OK. Maybe you can contribute this to https://github.com/MoonshotAI/kosong/blob/main/src/kosong/contrib/chat_provider/anthropic.py, adding a note comment stating that it's specific to Chutes Anthropic API. Thanks!

ZakWork:
Hey @stdrc 

Yes, the fix needs to be there, I have looked into this and there's a bit more to it. 

The reason for the crash was mainly due to the idea that for stream messages not all message types require input tokens to be included in the usage field, in fact this is only included in the first message "MessageStartEvent" the following events will not include this except for specific conditions. So most MessageDeltaEvent will have the input token field set as None in their MessageDeltaUsage. 

You can find more info on this here: 
https://docs.claude.com/en/docs/build-with-claude/streaming#full-http-stream-response

https://github.com/anthropics/anthropic-sdk-python/blob/d9aea38e754d55f8f0875fdf19ee44f78ca7b845/src/anthropic/types/raw_message_delta_event.py#L24

https://github.com/anthropics/anthropic-sdk-python/blob/d9aea38e754d55f8f0875fdf19ee44f78ca7b845/src/anthropic/types/message_delta_usage.py#L18

https://github.com/anthropics/anthropic-sdk-python/blob/d9aea38e754d55f8f0875fdf19ee44f78ca7b845/src/anthropic/types/usage.py#L23C4-L23C22

The current code is not set to handle this and therefore will fail not only for Chutes Anthropic API but most likely for all other providers using the Anthropic API given this is how the standard is suppose to work. 

This pr should fix the issue above and stop the Kimi CLI from crashing

https://github.com/MoonshotAI/kosong/pull/21


I have also identified a second issue with this and that is the MessageStartEvent is not used to updated the input token usage (this is used for context management and therefore this will also likely lead to an error). Furthermore, the usage update from MessageDeltaEvent is used to replace the usage object instead of updating it and although the data from MessageDeltaUsage is cumulative it does not necessarily have all the fields populated including the input tokens field. So currently the usage info comes in only from the MessageDeltaUsage and therefore the context limit is not calculated correctly (massively underestimated as the initial input tokens are discarded/not counted) which will probably result in an error from the provider for input being too long.  
 
I have a fix for this (make sure to use MessageStartEvent data to set the initial usage then update usage data from MessageDeltaUsage correctly) I can update the current pr above or open a new one?

I am a big fan of the work your team has done with Kimi K2 Thinking and I hope this helps the kimi cli get better support/adoption by the community, keep up the great work!

stdrc:
> I have also identified a second issue with this and that is the MessageStartEvent is not used to updated the input token usage (this is used for context management and therefore this will also likely lead to an error). Furthermore, the usage update from MessageDeltaEvent is used to replace the usage object instead of updating it and although the data from MessageDeltaUsage is cumulative it does not necessarily have all the fields populated including the input tokens field. So currently the usage info comes in only from the MessageDeltaUsage and therefore the context limit is not calculated correctly (massively underestimated as the initial input tokens are discarded/not counted) which will probably result in an error from the provider for input being too long.
> 
> I have a fix for this (make sure to use MessageStartEvent data to set the initial usage then update usage data from MessageDeltaUsage correctly) I can update the current pr above or open a new one?

Could you please directly push the fix of the second issue to https://github.com/MoonshotAI/kosong/pull/21? I guess it would be better to have them fixed together. Thanks for your time!

ZakWork:
That should be ready for you to review now.

PR #286, opened 2025-11-14
fix: put thinking content in AgentMessageChunk in ACP server

### LINKED ISSUE #282: Use Proper acp.schema.AgentThoughtChunk
### What version of Kimi CLI is running?

0.54

Kimi For Coding

kimi-for-coding

_No response_

`<think>` blocks in AgentMessageChunk on Zed

Just send "Hi" on Zed

Image

### Additional information

Use a function like:

```py
    async def _send_think(self, think: str):
        """Send thinking content to client."""
        if not self.session_id:
            return

        await self.connection.sessionUpdate(
            acp.SessionNotification(
                sessionId=self.session_id,
                update=acp.schema.AgentThoughtChunk(
                    content=acp.schema.TextContentBlock(type="text", text=think),
                    sessionUpdate="agent_thought_chunk",
                ),
            )
        )
```

And then in `_stream_events` function drop `run_state.in_thinking` and simply use:
```py
                case ThinkPart(think=think):
                    await self._send_think(think)
```

Iwoooooods:
## Related Issue

Fix #282 

Resolve #(issue_number)

## Description

thinking part are sent as normal text through ACP, which should be changed to acp.schema.AgentThoughtChunk to align with other models.
And removing two sent of thinking tags when switch thinking state passes all tests
image


yihong0618:
cc @PsiACE

src/kimi_cli/ui/acp/__init__.py
```diff
@@ -185,10 +185,8 @@ async def _stream_events(self, wire: WireUISide):
 
             assert self.run_state is not None
             if isinstance(msg, ThinkPart) and not self.run_state.in_thinking:
-                await self._send_text("<think>\n")
                 self.run_state.in_thinking = True
             if not isinstance(msg, ThinkPart) and self.run_state.in_thinking:
-                await self._send_text("\n\n</think>\n\n")
                 self.run_state.in_thinking = False
```
stdrc, line 190: Maybe we don't need to maintain `run_state.in_thinking` any more?

stdrc (commented):
Rest LGTM

stdrc:
Sorry mis-operated. Cherry-picked and reopened in https://github.com/MoonshotAI/kimi-cli/pull/313

PR #328, opened 2025-11-19
feat: introduce Kaos and LocalKaos for filesystem access

stdrc:
## Related Issue

Resolve #(issue_number)

### DIFF
--- a/src/kimi_cli/ui/shell/prompt.py
+++ b/src/kimi_cli/ui/shell/prompt.py
@@ -498,7 +498,7 @@ class CustomPromptSession:
             [
                 MetaCommandCompleter(),
                 # TODO(kaos): we need an async KaosFileMentionCompleter
-                LocalFileMentionCompleter(Path.cwd()),
+                LocalFileMentionCompleter(KaosPath.cwd().unsafe_to_local_path()),
             ],
             deduplicate=True,
         )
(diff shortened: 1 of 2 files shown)


src/kaos/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+        """Iterate over the entries in a directory."""
+        ...
+
+    def glob(
+        self, path: StrOrKaosPath, pattern: str, *, case_sensitive: bool = True
+    ) -> AsyncGenerator[KaosPath]:
+        """Search for files/directories matching a pattern in the given path."""
+        ...
+
+    async def readtext(
```
Epsirom, line 50: Why not `read_text`, just like `Path().read_text` ?

Epsirom:
Non-local KAOS implementation should be verified before merge. @powerfooI

src/kaos/__init__.py
stdrc: Seems the naming conventions of the functions in `os` module and `pathlib.Path` are not the same. So in `kaos` and in `KaosPath` I follow them separately.

powerfooI:
> Non-local KAOS implementation should be verified before merge. @powerfooI

Sure, but non-local implementation won't be merged here.

Issue #382, opened 2025-11-27
Shell mode crashes when DeepSeek (Anthropic format) omits usage input fields


leslieo2:
### What version of Kimi CLI is running?

0.58

### Which open platform/subscription were you using?

custom

### What issue are you seeing?

  Using a DeepSeek provider with Anthropic-style API responses, the LLM sometimes omits or returns null for usage.input_other (or other input fields). KimiSoul sums these
  fields directly, leading to None + int and a TypeError, causing the shell run to crash.

### What steps can reproduce the bug?

  1. Configure the model to a DeepSeek Anthropic endpoint (e.g., deepseek-chat).  type: anthropic base_url:https://api.deepseek.com/anthropic
  2. Run uv run kimi and enter any prompt (e.g., who are you).
  3. Observe TypeError: unsupported operand type(s) for +: 'NoneType' and 'int' pointing to TokenUsage.input.

  Expected Behavior
  If usage input fields are missing or null, treat them as zero and continue without crashing.

  Actual Behavior
  TokenUsage.input attempts to add None to int, raising a TypeError and terminating the session.

PR #383, opened 2025-11-27
fix: handle missing token usage fields to avoid shell crash (#382) 

leslieo2:
- Add _safe_usage to KimiSoul to tolerate None/missing usage fields when summing tokens.
  - Use the safe totals when updating context usage during step execution and result growth.
  - Add regression tests covering missing input tokens and full field aggregation.

### DIFF
--- a/src/kimi_cli/soul/kimisoul.py
+++ b/src/kimi_cli/soul/kimisoul.py
@@ -15,6 +15,7 @@ from kosong.chat_provider import (
     APIStatusError,
     APITimeoutError,
     ThinkingEffort,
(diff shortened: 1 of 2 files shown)


leslieo2:
fix: #382

stdrc:
I guess this can be fixed by https://github.com/MoonshotAI/kosong/pull/21?

leslieo2:
> I guess this can be fixed by [MoonshotAI/kosong#21](https://github.com/MoonshotAI/kosong/pull/21)?

yes, seems the same problem.

PR #401, opened 2025-11-28
fix: harden file tool path validation against shared-prefix escapes

leslieo2:
Problem: File tools validated paths with str.startswith, so a path like /work-evil/foo was treated as inside /work, allowing reads/writes outside the working directory.
Fix: Use canonical path containment checks for Read/Write/Replace/Glob and add regression tests to ensure shared-prefix escape attempts are rejected.

Tests before fix (expected failures)

  - On the vulnerable code, run the new regression cases and they fail because the tools allow shared-prefix escapes:
      - pytest tests/test_write_file.py::test_write_outside_work_directory_with_prefix
      - pytest tests/test_glob.py::test_glob_outside_work_directory_with_prefix
      - pytest tests/test_str_replace_file.py::test_replace_outside_work_directory_with_prefix


Eric-Guo:
I would rather allow Glob and Readfile be able to reading outside folder by default, because some rails app using gems which outside of the application folder....

leslieo2:
> I would rather allow Glob and Readfile be able to reading outside folder by default, because some rails app using gems which outside of the application folder....

We can relax Read/Glob later, but the immediate issue is that the old startswith guard was flawed—shared prefixes like /work vs /work-evil let escapes through. First we need a solid containment check to make the restriction actually work; then, if you want, we can add a controlled opt-in (e.g., allowlist or config toggle) for Read/Glob so you can access gems while keeping writes locked down and avoiding a security hole.

stdrc:
> I would rather allow Glob and Readfile be able to reading outside folder by default, because some rails app using gems which outside of the application folder....

Agree. Restriction on readonly tools should be relaxed.

PR #402, opened 2025-11-28
docs: no need move AGENTS.md to first message from system prompt

Eric-Guo:
## Related Issue

Tried at #293, found introduce too many code change

## Description

So no need move AGENTS.md to first message from system prompt


stdrc:
No hurry. I may do this some day...

PR #421, opened 2025-12-03
fix: ACP tool result streaming for Shell tool

comeonzhj:
## Related Issue

https://github.com/MoonshotAI/kimi-cli/issues/419

Resolve #(issue_number)

## Description

**Background**  
 When using Kimi CLI as an ACP agent inside Zed, I noticed that although the LLM does invoke the Shell (Bash) tool and the command actually runs, Zed’s tool card shows no output at all.  
  
 After debugging, it turned out that in the ACP UI adapter, only the `SetTodoList` tool result was propagated to the ACP client: for all other tools (including Shell/Bash), `ToolCallProgress` was sent without any `content`, so ACP clients only saw the tool status change, but never the actual output.  
  
 **What this change does**  
 1. In `src/kimi_cli/ui/acp/__init__.py`:  
    - Remove the special case that only sets `update.content` for `SetTodoList`, and instead always call `_tool_result_to_acp_content(tool_ret)` for **all** tools.  
    - Make `_tool_result_to_acp_content` more robust:  
      - It now handles `tool_ret.output` being a `str`, a `list[ContentPart]`, or a single `ContentPart`.  
      - If there is no output, it falls back to `tool_ret.message` so that ACP clients still receive a meaningful text payload instead of an entirely empty result.  
    - With this, in ACP mode the Shell/Bash tool’s stdout/stderr and other output are properly wrapped into `ToolCallProgress.content`, and ACP clients like Zed can render the command result as expected.  
 2. Update `CHANGELOG.md` under the `0.60` release with a “Fixed” entry:  
    - “ACP: Fix tool results (including Shell/Bash tool output) not being displayed in ACP clients like Zed”.  
  
 **How to verify**  
 - Environment: macOS, dependencies installed via `uv`, Kimi CLI used as an ACP agent in Zed.  
 - Locally:  
   - Run `uv run pytest tests/test_bash.py tests/test_tool_schemas.py -q` to ensure Shell tool behavior and tool schemas are still valid.  
   - Start the ACP server with:  
     ```bash
     uv run python -m kimi_cli.cli --acp
     ```  
   - Configure Zed’s `settings.json` like:  
     ```json
     "agent_servers": {
       "Kimi CLI Dev": {
         "command": "uv",
         "args": ["run", "python", "-m", "kimi_cli.cli", "--acp"],
         "working_directory": "/YOUR/PATH/TO/kimi-cli"
       }
     }
     ```  
   - Create a “Kimi CLI Dev” thread in Zed, ask it to run a Shell command (e.g. “please run \`ls\`”), approve the permission request, and confirm that the tool card now shows the actual Shell/Bash output.  
  
 **Impact**  
 - The change only affects the ACP UI adapter; shell mode, print mode and wire mode are untouched.  
 - Tool results in ACP mode are now consistently visible for all tools, not just `SetTodoList`.  
 - Shell/Bash behavior itself is unchanged; this PR only fixes how its result is surfaced to ACP clients.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Internal builds may append content to the Unreleased section.
 Only write entries that are worth mentioning to users.
 -->
 
+## [Unreleased]
+
+- ACP: Fix tool results (including Shell tool output) not being displayed in ACP clients like Zed
+
 ## [0.61] - 2025-12-04
 
 - Lib: Fix logging when used as a library
@@ -18,7 +22,6 @@ Only write entries that are worth mentioning to users.
 ## [0.60] - 2025-12-01
 
 - LLM: Fix interleaved thinking for Kimi and OpenAI-compatible providers
-- ACP: Fix tool results (including Shell tool output) not being displayed in ACP clients like Zed
 
 ## [0.59] - 2025-11-28
 


stdrc:
Shall we just fix it for all tools?

comeonzhj:
修改已经作用到所有工具了。此刻所有工具统一调用 _tool_result_to_acp_content，这样所有工具的结果都会在 ACP 客户端中显示出来，而不仅仅是 SetTodoList 或 Shell/Bash。

---
The changes have been applied to all tools. All tools now call _tool_result_to_acp_content uniformly, so that the results of all tools are displayed in the ACP client, not just SetTodoList or Shell/Bash.

stdrc:
可以，麻烦修一下 check

---
Okay, please fix it check

PR #438, opened 2025-12-06
Feat/symlink security

luojiyin1987:
## Related Issue

Resolve #(issue_number)


stdrc:
Why do we need to protect against symlink that points to something outside the workspace?

luojiyin1987:
We sandbox all file tools to the workspace. A symlink inside the workspace can point to /etc/passwd or ~/.ssh; without the resolve check we’d happily read/write those, which is an obvious local info-leak/overwrite risk.

If you think this isn't a problem, you can turn off pr.

PR #472, opened 2025-12-13
feat(acp): add `agent_info` to initialization response

RayZhao1998:
## Motivation

According to the [ACP Initialization](https://agentclientprotocol.com/protocol/initialization) specification, the agent SHOULD provide a name and version to the Client during initialization.

> It SHOULD also provide a name and version to the Client as well.

## What this PR does

This PR adds an agent_info field to the ACP initialization response, populated with the agent's name and version derived from the package metadata.

### DIFF
--- a/src/kimi_cli/constant.py
+++ b/src/kimi_cli/constant.py
@@ -2,6 +2,6 @@ from __future__ import annotations
 
 import importlib.metadata
 
-NAME = importlib.metadata.metadata("kimi-cli")["Name"]
+NAME = "Kimi CLI"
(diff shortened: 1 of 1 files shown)


src/kimi_cli/constant.py
```diff
@@ -2,5 +2,6 @@
 
 import importlib.metadata
 
+NAME = importlib.metadata.metadata("kimi-cli")["Name"]
```
stdrc:
```suggestion
NAME = "Kimi CLI"
```

PR #480, opened 2025-12-15
refactor: use ToolReturnValue for ToolResultBuilder

xxchan:
also add an `extras` method

## Related Issue

Resolve #(issue_number)


xxchan:
Need to fix tests

Issue #444, opened 2025-12-08
Models Switch on the fly


DesignerEE:
### What feature would you like to see?

Make option to switch models with TAB key, while asking questions. Same feature like OpenCode has.

Sepush:
It looks like the current `kimi-cli` doesn’t support a quick switch command like `/model.`
Right now, I have to go through these steps to switch models, which is quite inconvenient:

1. `/setup`
2. Select API platform and enter the key
3. Select the model

stdrc:
Now Kimi CLI supports `/model` to switch models on the fly

PR #502, opened 2025-12-18
feat: add /model meta command to select model

RayZhao1998:
## Related Issue

#444

## Motivation

Currently, switching models in Kimi CLI requires running `/setup`, which forces users to re enter the API key each time. This makes model switching unnecessarily heavy during daily usage.

This PR introduces a new `/model` meta command that allows users to quickly switch between already configured models without re running setup.

### 1. Persist all models during setup

Previously, `setup` only stored the user selected model in the config.
This PR changes the behavior to:

* Save all discovered models into `config.models`

This ensures the model list is available for later selection.

### 2. Add `/model` meta command

* Introduces a new `/model` meta command
* Displays models already recorded in the config, default and current active model
* Allows interactive selection of a model
* Switches the underlying runtime LLM immediately for the current session

## Behavior to Be Aware Of

* If the current model supports `thinking` and thinking mode is enabled, then the user switches to a model that does not support `thinking` via `/model`:

  * thinking mode will be automatically disabled
  * a toast notification will be shown to inform the user about this change

## Known Issue

1. If the current session has already used certain model capabilities, such as image input, switching to a model that does not support those capabilities may cause subsequent requests to fail.

    At the moment, all available Kimi models do not support image or similar advanced capabilities, it can be tracked separately in a follow up issue if needed.

### DIFF
--- a/src/kimi_cli/ui/shell/prompt.py
+++ b/src/kimi_cli/ui/shell/prompt.py
@@ -709,6 +709,8 @@ class CustomPromptSession:
         return True
 
     async def prompt(self) -> UserInput:
+        self._ensure_thinking_supported()
+
         with patch_stdout(raw=True):
             command = str(await self._session.prompt_async()).strip()
             command = command.replace("\x00", "")  # just in case null bytes are somehow inserted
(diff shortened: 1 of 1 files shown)


stdrc:
Supported in https://github.com/MoonshotAI/kimi-cli/pull/556

PR #562, opened 2026-01-01
feat(llm): support local llm (lm_studio)

YoungY620:
## Related Issue

N/A

### Motivation

Add support for locally deployed models to enable more users to try Kimi CLI. LM Studio was chosen for its ease of use and broad accessibility.

(To be honest, I really need it. 🆘 I found no way but to change the code.)

### Changes

1. **New `lm_studio` provider** (`llm.py`)
   
   Added dedicated provider for LM Studio. A separate provider was needed due to `httpx` compatibility issues (requires explicit `AsyncHTTPTransport` to work with LM Studio).

2. **LM Studio option in `/setup`** (`setup.py`)
   
   Added "LM Studio (Local)" to the setup wizard with auto-detection of loaded models.

### Usage

Run `/setup` and select "LM Studio (Local)", or configure manually:

```
[providers.local]
type = "lm_studio"
base_url = "http://localhost:1234/v1"
api_key = "any"  # LM Studio does not validate API key, any non-empty value works

[models.my-model]
provider = "local"
model = "model-name"
max_context_size = 32768
```

### DIFF
--- a/src/kimi_cli/ui/shell/setup.py
+++ b/src/kimi_cli/ui/shell/setup.py
@@ -32,7 +32,7 @@ if TYPE_CHECKING:
 class _Platform(NamedTuple):
     id: str
     name: str
-    base_url: str
+    default_base_url: str
     search_url: str | None = None
     fetch_url: str | None = None
     allowed_prefixes: list[str] | None = None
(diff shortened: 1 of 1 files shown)


stdrc (dismissed):
Good work on adding LM Studio support! The implementation is clean and user-friendly.

**Highlights:**
1. Separate provider type (`lm_studio`) with proper httpx transport handling
2. Nice setup wizard flow with auto-detection of loaded models
3. Sensible defaults (128k context, localhost:1234)
4. Context size validation with minimum check

**Minor suggestions (optional):**
1. Consider adding a note in the setup about why LM Studio needs its own provider (httpx compatibility)
2. The hardcoded `api_key=SecretStr("lm-studio")` works but could be any non-empty value as mentioned in docs

**Testing note:** Might be worth documenting how to test this locally for other reviewers.

LGTM overall!

Reviewed by Alice (AI agent)

PR #566, opened 2026-01-04
fix: do not run this on fork

yihong0618:
## Related Issue

Resolve #(issue_number)

## Description

as is: https://github.com/yihong0618/kimi-cli/actions/runs/20689892393


stdrc:
Simple and necessary fix! 

The `if: github.repository == 'MoonshotAI/kimi-cli'` condition is the standard way to prevent workflows from failing on forks due to missing secrets/permissions.

LGTM! 👍

PR #568, opened 2026-01-04
feat: add ralph loop

xxchan:
## Summary
- add max_ralph_iterations to loop control and CLI override
- implement Ralph loop stop handling and safeword injection per turn
- render inline XML/HTML tags as plain text in shell markdown
- update config validation and snapshots

## Testing
- not run (not requested)

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -60,15 +60,16 @@ Term mapping (Chinese <-> English, and proper noun handling):
 | Agent Skills | Agent Skills | yes | yes |
 | Skill | skill | yes | no |
 | 系统提示词 | system prompt | no | no |
+| 提示词 | prompt | no | no |
 | 会话 | session | no | no |
 | 上下文 | context | no | no |
 | 子 Agent | subagent | yes (Agent) | no |
-| API 密钥 | API key | no | no |
-| JSON | JSON | no | no |
-| JSONL | JSONL | no | no |
-| OAuth | OAuth | no | no |
-| macOS | macOS | no | no |
-| uv | uv | no | no |
+| API 密钥 | API key | yes | no |
+| JSON | JSON | yes | yes |
+| JSONL | JSONL | yes | yes |
+| OAuth | OAuth | yes | yes |
+| macOS | macOS | yes | yes |
+| uv | uv | yes | yes |
 | 审批请求 | approval request | no | no |
 | 斜杠命令 | slash command | no | no |
 | 工具调用 | tool call | no | no |
--- a/docs/en/configuration/config-files.md
+++ b/docs/en/configuration/config-files.md
@@ -47,7 +47,7 @@ model = "kimi-for-coding"
 max_context_size = 262144
 
 [loop_control]
-max_steps_per_run = 100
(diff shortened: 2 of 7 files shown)


stdrc (commented):
Reviewed the Ralph loop implementation. Overall looks good!

**Summary:**
- Adds `max_ralph_iterations` config option and CLI flag
- Implements safeword mechanism (`<safeword>STOP</safeword>`) for agent-controlled loop exit
- Stops on tool rejection as well
- Good test coverage for different scenarios

**Observations:**
1. The documentation is well-written in both EN and ZH
2. Config validation correctly rejects values < -1
3. Test cases cover: iteration limit, safeword stop, tool rejection, and disabled state

**Question:**
- Is there a plan to add a timeout mechanism in addition to iteration count? Long-running loops could benefit from a time-based cutoff.

Overall: Clean implementation of a useful feature.

Reviewed by Alice (AI agent)

xxchan:
Thanks @stdrc for the review! On the timeout question: not in this PR yet. Right now the only guardrail is `max_ralph_iterations` (incl. `-1` to disable). If we see a need for a time-based cutoff, I’m happy to add a follow-up config (e.g., max duration) in a separate PR.

src/kimi_cli/cli/__init__.py
```diff
@@ -372,6 +384,14 @@ def kimi(
...
         config = config_file
 
+    if max_ralph_iterations is not None:
+        if isinstance(config, Config):
+            config.loop_control.max_ralph_iterations = max_ralph_iterations
+        else:
+            config_path = config if isinstance(config, Path) else None
+            config = load_config(config_path)
+            config.loop_control.max_ralph_iterations = max_ralph_iterations
+
```
stdrc: This will cause the `/reload` slash command (and `/setup`) failed to reload config from the default config file location.

src/kimi_cli/utils/rich/markdown.py
```diff
@@ -775,7 +775,8 @@ def __rich_console__(self, console: Console, options: ConsoleOptions) -> RenderR
             exiting = token.nesting == -1
             self_closing = token.nesting == 0
 
-            if node_type == "text":
+            if node_type in {"text", "html_inline", "html_block"}:
+                # Render HTML tokens as plain text so safeword markup stays visible.
```
stdrc, line 779: What's this?

src/kimi_cli/cli/__init__.py
xxchan: Good catch - the CLI override was materializing the config object, so /reload stopped re-reading the default config path. I moved the max_ralph_iterations override into KimiCLI.create (after load) and keep CLI config as a path/None, so reload now reads from disk then applies the override. Thanks!

src/kimi_cli/utils/rich/markdown.py
xxchan: This renders html_inline/html_block as plain text so the <safeword>STOP</safeword> markup stays visible; otherwise markdown HTML handling can hide or treat it specially. It is specifically for Ralph-mode safeword visibility.

PR #572, opened 2026-01-05
fix(tools): expand ~ path in ReadFile tool

xpzouying:
## Summary
- Expand `~` to home directory in ReadFile tool using `Path.expanduser()`
- Add unit test for tilde path expansion

## Background
When reading `~/.local_rc` from working directory `~/cc`, the tool failed because `~` was not expanded. The path was resolved to `~/cc/~/.local_rc` instead of `/home/user/.local_rc`.

## Test plan
- [x] All existing tests pass (`pytest tests/test_read_file.py -v`)
- [x] New test `test_read_with_tilde_path_expansion` passes
- [x] `prek run --all-files` passes


stdrc (approved):
LGTM! Clean and straightforward fix.

The change correctly uses `Path.expanduser()` to handle tilde expansion before creating the KaosPath. The test is well-written and includes proper cleanup.

Minor suggestion (optional): You could also handle `~username` syntax if needed in the future, but `expanduser()` already supports this, so the current implementation should work.

Reviewed by Alice (AI agent)

stdrc:
Great fix! The implementation looks clean and correct:

✅ Uses standard `Path.expanduser()` for tilde expansion  
✅ Well-documented PR with clear background explanation  
✅ Comprehensive unit test that creates/cleans up test file properly  
✅ Follows existing code patterns

LGTM! 👍

---
🤖 Review by Bob (AI Agent)

stdrc:
**Update: Codex Review Finding**

[P2] The test writes artifacts into the real home directory, which is risky:

```python
# tests/test_read_file.py:327-349
home = Path.home()
test_file = home / ".kimi_test_expanduser_temp"
```

**Issues:**
- If that file already exists, it could be overwritten or deleted
- Permission failures in restricted environments
- Breaks test isolation principles

**Suggestion:** Consider monkeypatching `HOME`/`Path.home()` to a temporary directory, or use pytest's `tmp_path` fixture to simulate `~` expansion without touching the real home.

Example approach:
```python
async def test_read_with_tilde_path_expansion(read_file_tool: ReadFile, tmp_path: Path, monkeypatch):
    # Redirect home to temp directory
    monkeypatch.setenv('HOME', str(tmp_path))
    monkeypatch.setattr(Path, 'home', lambda: tmp_path)
    
    test_file = tmp_path / '.kimi_test_expanduser_temp'
    test_file.write_text('Test content')
    
    result = await read_file_tool(Params(path='~/.kimi_test_expanduser_temp'))
    # ...
```

stdrc:
**Additional Architectural Feedback**

There's a deeper issue: the fix uses `Path.expanduser()` instead of working within the `KaosPath` abstraction.

```python
# Current implementation
expanded_path = Path(params.path).expanduser()  # Uses stdlib Path
p = KaosPath(str(expanded_path))
```

**Problem:** `KaosPath` is designed to potentially operate on remote nodes (via Kaos). Using `Path.expanduser()` expands `~` on the **local** machine, which would be incorrect if the path refers to a remote file.

**Suggestion:** The tilde expansion should be handled within `KaosPath` itself, so it works correctly for both local and remote paths. Consider:
1. Adding an `expanduser()` method to `KaosPath`
2. Or handling this at a higher level with awareness of whether the path is local vs remote

stdrc (changes requested):
需要重新审视架构问题：应该在 KaosPath 层面处理 ~ 展开，而非使用 Path.expanduser()

stdrc (approved):

PR #573, opened 2026-01-05
fix: fix the dependency conflict between kimi-cli/pykaos and gradio

pvzheroes125:
degrade the version of `aiofiles` from **25.1.0** to **24.0~25.0** in order to resolve the dependency conflict between `gradio` and `kimi-cli/pykaos`

## Related Issue

Resolve #(issue_number)

### DIFF
--- a/packages/kaos/pyproject.toml
+++ b/packages/kaos/pyproject.toml
@@ -5,7 +5,7 @@ description = ""
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "aiofiles==25.1.0",
+    "aiofiles>=24.0,<25.0",
     "asyncssh==2.21.1",
 ]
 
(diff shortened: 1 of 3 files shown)


pvzheroes125:
@stdrc

stdrc:
Thanks for the fix! A few questions:

1. Could you link the related issue or provide more details about the specific dependency conflict? What error message were you seeing?

2. Is there a specific version of gradio that requires aiofiles < 25.0? It would be helpful to document this constraint.

3. Have you tested that the downgrade doesn't break any existing functionality?

pvzheroes125:
@stdrc 
1. 
img_v3_02tl_22278a18-9678-435f-8c3a-efa19fdc5c9g
2. all versions of gradio by now requires aiofiles < 25.0
https://github.com/gradio-app/gradio/blob/main/requirements.txt#L1
3. Yes, kimi-cli/pykaos now does not necessarily need aiofiles==25.1.0.

PR #615, opened 2026-01-14
proposal: add Init method and ExternalToolCall request

x5iu:


stdrc:
close in favor of #651

PR #639, opened 2026-01-18
feat(config): add reserved_context_size setting for configurable compaction trigger

### LINKED ISSUE #351: Provide a configuration setting to adjust threshold for auto-compact trigger.

Different kinds of tasks require different level of model accuracy and context decay is a well-known phenomenon. Currently the auto-compact in kimi-cli triggers at around 80%. It'd be great if users could adjust this value themselves while still having a sensible default value (eg 80%) in place.

### Additional information

_No response_

zerone0x:
## Related Issue

Resolve #351

## Description

This PR adds a configurable `auto_compact_threshold` setting that allows users to adjust when auto-compaction triggers.

### Changes:
- Added `auto_compact_threshold` field to `LoopControl` config with default value 0.8 (80%)
- Valid range is 0.1-1.0 (10%-100%)
- Updated `KimiSoul` to use percentage-based threshold instead of the previous fixed reserved_tokens approach
- Added tests for the new configuration option including boundary validation

### Usage:
Users can configure the threshold in their `config.toml`:

```toml
[loop_control]
auto_compact_threshold = 0.7  # trigger compaction at 70% context usage
```

This addresses the user request to be able to adjust the auto-compact trigger point based on their task requirements.

## Checklist

🤖 Generated with Claude Code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Only write entries that are worth mentioning to users.
 
 ## Unreleased
 
+- Config: Add `reserved_context_size` setting to customize auto-compaction trigger threshold (default: 50000 tokens)
+
 ## 0.81 (2026-01-21)
 
 - Skills: Add flow skill type with embedded Agent Flow (Mermaid/D2) in SKILL.md, invoked via `/flow:<skill-name>` commands
--- a/docs/en/configuration/config-files.md
+++ b/docs/en/configuration/config-files.md
@@ -52,6 +52,7 @@ max_context_size = 262144
 max_steps_per_turn = 100
 max_retries_per_step = 3
 max_ralph_iterations = 0
+reserved_context_size = 50000
 
 [services.moonshot_search]
 base_url = "https://api.kimi.com/coding/v1/search"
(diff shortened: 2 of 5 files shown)


stdrc:
What about `reserved_context_size`? IMO what we really care about in compaction is not the percentage, it's the remained token count.

zerone0x:
Thanks for the feedback! You're right - the reserved token count is more important than the percentage, since the LLM needs a consistent amount of space for generation regardless of the total context size.

I can update the PR to use `reserved_context_size` instead:

```python
class LoopControl(BaseModel):
    # ...
    reserved_context_size: int = Field(default=50_000, ge=1000)
    """Reserved token count that triggers auto-compaction. Default is 50000."""
```

And the compaction trigger would be:
```python
if self._context.token_count + self._loop_control.reserved_context_size >= self._runtime.llm.max_context_size:
```

This keeps the original logic but makes the reserved token count configurable. Should I push this change?

stdrc:
> Should I push this change?

Yeah sure. That would be sweet. Thank you!

zerone0x:
@stdrc Done! Pushed commit 8d0bd0b with the changes:

- Renamed `auto_compact_threshold` to `reserved_context_size` (default: 50000 tokens)
- Restored the original token-based compaction logic: `context_tokens + reserved_context_size >= max_context_size`
- Updated tests accordingly

The config usage is now:
```toml
[loop_control]
reserved_context_size = 50000  # can adjust based on needs
```

PR #650, opened 2026-01-20
feat(shell): add /timeout slash command for configurable shell timeout

### LINKED ISSUE #625: Add /timeout so we can adjust the timeout

When using kimi cli to install something then it keeps getting killed by timeout (60s) if we could spesify our timeout for our task becasue different tasks neeed different ammount of time, it would be perfect.

Image

### Additional information

_No response_

zerone0x:
## Related Issue

Fixes #625

## Description

Add the ability to configure default shell command timeout via:
- New `shell.default_timeout` config option (1-300 seconds, default: 60)
- New `/timeout` slash command to view/set the default timeout
- Shell tool now uses configured timeout as default instead of hardcoded 60s

This addresses user feedback about long-running commands being killed by the default 60-second timeout. Users can now set a longer default with `/timeout 300` without having to tell the LLM each time.

### Usage

```bash
# View current timeout
/timeout
# Current default shell timeout: 60 seconds
# Usage: /timeout <seconds> (1-300)

# Set new timeout
/timeout 300
# Default shell timeout set to 300 seconds. Reloading...
```

### Config

The timeout can also be set directly in `config.toml`:

```toml
[shell]
default_timeout = 300
```

## Changes

- `src/kimi_cli/config.py`: Add `ShellConfig` with `default_timeout` field
- `src/kimi_cli/tools/shell/__init__.py`: Create dynamic params class with configured default timeout
- `src/kimi_cli/ui/shell/slash.py`: Add `/timeout` slash command
- `tests/conftest.py`: Update shell_tool fixture to pass config
- `tests/test_config.py`: Update config snapshot test

## Checklist

---
🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/kimi_cli/config.py
+++ b/src/kimi_cli/config.py
@@ -111,6 +111,13 @@ class MCPConfig(BaseModel):
     )
 
 
+class ShellConfig(BaseModel):
+    """Shell tool configuration."""
+
+    default_timeout: int = Field(default=60, ge=1, le=300)
+    """Default timeout in seconds for shell commands (1-300). LLM can override per-call."""
+
+
 class Config(BaseModel):
     """Main configuration structure."""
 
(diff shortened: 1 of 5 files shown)


xxchan:
According to https://github.com/MoonshotAI/kimi-cli/issues/625#issuecomment-3758028288, this seems not needed.

Please discuss in the existing issue before a PR, thanks. https://github.com/MoonshotAI/kimi-cli/blob/main/CONTRIBUTING.md

PR #700, opened 2026-01-25
fix(tools): improve ReadFile description to clarify directory handling

xpzouying:
AI 模型有时会用 ReadFile 读取目录路径，导致 Invalid path 错误。

在工具描述开头添加明确声明：只能读取文件，不能读取目录。

---
AI models sometimes use ReadFile to read directory paths, causing Invalid path errors.

Add an explicit statement at the beginning of the tool description: only files can be read, not directories.


xpzouying:
image

xxchan:
请问你使用的模型是？

---
What model are you using?

PR #739, opened 2026-01-28
fix(kosong): strip JSON Schema metadata from Google GenAI tool parameters

### LINKED ISSUE #734: [Bug]: Google GenAI provider fails with extra_forbidden for tool parameters containing $schema

1.2

google_genai

gemini-3-pro-preview

Linux

When using Exa MCP (`https://mcp.exa.ai/mcp`) or other tools that provide standard JSON Schema parameters, a validation error occurs during the tool calling phase:

```text
Unexpected error: 1 validation error for FunctionDeclaration
parameters.$schema
  Extra inputs are not permitted [type=extra_forbidden, input_value='http://json-schema.org/draft-07/schema#', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
```

This error is caused by the `google-genai` SDK's Pydantic model enforcing strict validation on the `FunctionDeclaration` parameters field, rejecting the `$schema` metadata field.

1. Configure the `google_genai` provider.
2. Add Exa MCP: `kimi mcp add exa https://mcp.exa.ai/mcp`.
3. Execute any command triggering a search (e.g., `Search for latest AI news using exa`).
4. The error occurs.

**Minimal Reproduction Code (using `uv`):**
```bash
uv run --no-project --python 3.12 --with google-genai --with pydantic python -c "
from google.genai.types import FunctionDeclaration
try:
    FunctionDeclaration(
        name='test',
        description='test',
        parameters={'$schema': 'http://json-schema.org/draft-07/schema#', 'type': 'object'}
    )
except Exception as e:
    print(e)
"
```

Tool calls should proceed successfully. The provider should strip unsupported JSON Schema keywords (like `$schema`) from the parameters before passing them to the `google-genai` SDK.

xiaoju111a:
## Related Issue

Resolves #734

## Description

This PR fixes a compatibility issue between Google GenAI provider and MCP tools that include standard JSON Schema metadata fields.

### Problem

When using MCP tools (like Exa MCP) with Google GenAI provider, the following validation error occurs:

```
1 validation error for FunctionDeclaration
parameters.$schema
  Extra inputs are not permitted [type=extra_forbidden, input_value='http://json-schema.org/draft-07/schema#', input_type=str]
```

**Root cause**: 
- MCP tools' `inputSchema` includes standard JSON Schema metadata fields
- Google GenAI SDK's Pydantic model has `extra='forbid'`, which rejects these additional fields
- Kimi CLI was passing the complete schema without filtering metadata

### Solution

Strip JSON Schema metadata fields in `tool_to_google_genai()` function before passing to Google GenAI SDK. We filter 4 fields that are rejected by the SDK:

- `$schema`, `$id`, `$comment` - JSON Schema metadata
- `examples` - Example values (not part of validation schema)

Note: `$defs` and `definitions` are already removed by kosong's `deref_json_schema()` function, so we don't need to filter them here.

```python
def tool_to_google_genai(tool: KosongTool) -> Tool:
    # Strip JSON Schema metadata fields (google-genai SDK has extra='forbid')
    # Note: $defs/definitions are already removed by kosong's deref_json_schema()
    parameters = {
        k: v
        for k, v in tool.parameters.items()
        if k not in ("$schema", "$id", "$comment", "examples")
    }
    
    return Tool(
        function_declarations=[
            FunctionDeclaration(
                name=tool.name,
                description=tool.description,
                parameters=parameters,
            )
        ]
    )
```

### Impact

- ✅ Only affects Google GenAI provider
- ✅ Backward compatible (tools without metadata work as before)
- ✅ No performance impact (simple dict filtering)
- ✅ Follows JSON Schema best practices (metadata fields are not parameter definitions)

### Testing

Manually verified the 4 rejected fields with Google GenAI SDK:

```python
# Tested fields that cause ValidationError:
# ❌ $schema, $id, $comment, examples

# After fix: ✅ All metadata stripped automatically
# Tools work correctly with Google GenAI provider
```

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works.

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/739">
  <picture>
    Open with Devin
  </picture>
</a>


percymcn:
There is a one-line version of this that removes the whole class of failure instead of the four keys, and I think it's worth considering before this lands: **pass the schema to `parameters_json_schema=` rather than `parameters=`.**

`FunctionDeclaration` has both fields. `parameters=` is the narrow OpenAPI-style `Schema` proto (`extra='forbid'`, which is what raises here); `parameters_json_schema=` is the JSON-Schema field, and it takes the schema **verbatim** — no stripping, no rewriting. Measured on `google-genai==2.17.0` with an Exa-shaped schema carrying every keyword at once:

```python
RAW = {"$schema": "http://json-schema.org/draft-07/schema#", "type": "object",
       "properties": {"a": {"type": ["string", "null"]},
                      "c": {"type": "integer", "exclusiveMinimum": 0},
                      "d": {"type": "array", "prefixItems": [{"type": "integer"}]}}}

FunctionDeclaration(name="t", parameters=RAW)              # 6 validation errors
FunctionDeclaration(name="t", parameters_json_schema=RAW)  # accepted, byte-identical
```

This isn't an obscure corner — it's what the JS client already does for you. `@google/genai` checks for a top-level `$schema` and, when it finds one, routes the declaration to `parametersJsonSchema` and sends it unmodified. So the key that's crashing the Python client is the same key that tells the JS client to use the field that accepts everything. Python has no such auto-routing (there is no `$schema` handling anywhere in the package), which is why this only shows up here.

Two things about the current patch that I'd want on the record either way, both measured against the diff as written:

**1. The strip is top-level only.** `tool.parameters.items()` isn't recursive, and `$comment` / `examples` / `$id` almost always appear on individual properties rather than at the root. Running the PR's exact transform:

```
PR-fixed, top-level $schema     OK          <- the reported Exa case
PR-fixed, nested $comment       REJECTED    parameters.properties.q.$comment
PR-fixed, nested examples       REJECTED    parameters.properties.q.examples
PR-fixed, nested $id            REJECTED    parameters.properties.q.$id
```

**2. Even a recursive strip leaves about a dozen more, and most of them have no `$`.** Same transform, one keyword per run:

```
type: ["string","null"]  REJECTED    prefixItems      REJECTED
exclusiveMinimum         REJECTED    uniqueItems      REJECTED
enum: [1, 2]             REJECTED    const            REJECTED
oneOf                    REJECTED    anyOf            OK  (control)
```

plus array-form `items`, `patternProperties`, `propertyNames`, `allOf` and `not`. A `$`-prefix heuristic catches none of these, and MCP servers emit them routinely.

**The two that matter most can't be fixed by stripping at all**, because deleting them destroys the parameter rather than sanitising it:

- `type: ["string", "null"]` — the standard spelling for an optional parameter. Dropping `type` leaves the model with no idea what to send. The lossless rewrite is `{"type": "string", "nullable": true}` — `nullable` is a real `Schema` field, verified accepted. (This is also exactly what `@google/genai` does silently on the narrow path.)
- `prefixItems: [int, int]` / array-form `items` — dropping it turns "exactly two integers" into "an array of anything, any length", which returns a 200 and garbage. Lossless when the positions share a type: `items: {...}` + `minItems`/`maxItems`, both accepted.

So if the strip approach is preferred over the field switch, it needs to be a recursive **rewrite**, not a recursive delete.

Last thing, and it makes any of these testable: `FunctionDeclaration(...)` raises offline with no API key and no network, so the repro in #734 is already a CI test. Iterating a table of keyword → schema through it pins the accepted set in a few seconds and catches the next SDK version that moves the boundary.

(One surface caveat if this pattern gets reused elsewhere: the same SDK **does** inline `$ref`/`$defs` for you on `response_schema`, but not on `FunctionDeclaration.parameters`, where a `$ref` is rejected outright. Doesn't affect this PR — `deref_json_schema()` has already run — but the behaviour isn't uniform across the SDK's surfaces.)

PR #888, opened 2026-02-04
fix(kosong): handle string annotations in SimpleToolset return type check

0xWelt:
When using `from __future__ import annotations`, type annotations are stored as strings rather than actual types.

## Changes

1. Uses `inspect.signature().return_annotation` to get the raw annotation
2. Handles string annotations by checking if it matches any suffix of `kosong.tooling.ToolReturnValue` (e.g., `ToolReturnValue`, `tooling.ToolReturnValue`, `kosong.tooling.ToolReturnValue`)
3. Falls back to strict type identity check for non-string annotations

## Tests

Added 3 new test cases in `test_tool_call.py`:
- `test_simple_toolset_with_string_annotations_callable_tool`
- `test_simple_toolset_with_string_annotations_callable_tool2`
- `test_simple_toolset_with_string_annotations_handle`


0xWelt:
Replaced by new PR with clean rebase on updated main

PR #1393, opened 2026-03-10
fix(acp): route shell commands through terminal args

hanhan3344:
## Summary
- fix ACP Shell terminal execution to pass the shell executable in `command` and the shell invocation in `args`
- adapt the ACP terminal integration to the current ACP SDK response shape using `terminal_id`
- add a regression test covering bash and PowerShell command/args routing

## Problem
When `kimi acp` was used behind ACP clients with terminal support, Shell commands like `echo test` could fail with errors like:

- `spawn echo test ENOENT`
- `spawn ls -la ENOENT`

The ACP terminal adapter passed the entire shell command string as `terminal/create.command`, which made clients try to spawn the full string as an executable path.

The adapter also assumed `create_terminal()` returned a `TerminalHandle`, but the current ACP SDK returns a response containing `terminal_id`, with follow-up terminal operations happening through separate RPC methods.

## Fix
- use the existing Shell tool's `_shell_args()` to split shell execution into executable + args
- call `terminal/create` with:
  - `command=<shell executable>`
  - `args=[shell flags..., user command]`
- use `terminal_id` with `terminal/wait_for_exit`, `terminal/output`, `terminal/kill`, and `terminal/release`

## Validation
- `uv run pytest tests/acp/test_protocol_v1.py tests/acp/test_terminal_tool.py -q`

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/1393" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/kimi_cli/acp/tools.py
+++ b/src/kimi_cli/acp/tools.py
@@ -56,6 +56,7 @@ class Terminal(CallableTool2[ShellParams]):
         # Use the `name`, `description`, and `params` from the existing Shell tool,
         # so that when this is added to the toolset, it replaces the original Shell tool.
         super().__init__(shell_tool.name, shell_tool.description, shell_tool.params)
+        self._shell_tool = shell_tool
         self._acp_conn = acp_conn
         self._acp_session_id = acp_session_id
         self._approval = approval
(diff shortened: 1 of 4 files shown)


hanhan3344:
Addressed Copilot's maintainability suggestion in follow-up commit `e1d00c7`:

- exposed a public `Shell.shell_args()` helper
- updated the ACP terminal adapter to call the public API instead of the private `_shell_args()` method
- added a focused unit test for `Shell.shell_args()`

Validation:
`uv run pytest tests/acp/test_protocol_v1.py tests/acp/test_terminal_tool.py tests/tools/test_shell_invocation.py -q`

PR #1403, opened 2026-03-11
fix(shell): scope file-mention walk to typed directory prefix

### LINKED ISSUE #1375: File mentions (@) can't find the files

1.18.0

Kimi Code

kimi-for-coding

Darwin 24.6.0 arm64 arm

Since Kimi CLI 1.18.0, file mention suggestions are broken, most of the files cannot be found at all. Only files inside the current directory are suggested.

Run kimi cli in a big directory with lots of files and subfolder tree full of files.

It seems that this happens in big directory trees, not with very simple folder structures.

The CLI suggests the file I'm trying to reference.

### Additional information

The CLI only suggests files inside the current directory.

howardpen9:
## Summary

Fixes #1375 — In large repositories, the `@` file-mention completer becomes unable to find files in directories that sort alphabetically late (e.g., `src/`, `tests/`). The root cause is that `_get_deep_paths()` walks the entire project tree from the root and stops at `limit` (default 1000), so early directories exhaust the budget before later ones are ever visited.

### Changes

- **Scope the walk**: When the typed fragment contains a `/` (e.g., `@src/utils/hel`), extract the directory prefix and start `os.walk()` from that subdirectory instead of the project root. This ensures completions always search the intended subtree regardless of repo size.
- **Scope-aware caching**: The path cache now tracks which scope it was built for (`_cache_scope`), so switching between directories correctly invalidates stale results.
- **Include scope directory**: The scoped directory itself (e.g., `src/`) is included in results so existing behavior (showing the directory entry) is preserved.

### Test plan

- [x] Added `test_scoped_walk_finds_deep_files_in_large_repos` — regression test with 15 early directories exhausting a small limit, verifying late directories are still reachable
- [x] Added `test_scoped_walk_with_partial_filename` — verifies scoped walk works with partial filename after slash
- [x] All 8 existing + new tests pass

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/1403" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/kimi_cli/ui/shell/prompt.py
+++ b/src/kimi_cli/ui/shell/prompt.py
@@ -225,6 +225,7 @@ class LocalFileMentionCompleter(Completer):
         self._limit = limit
         self._cache_time: float = 0.0
         self._cached_paths: list[str] = []
+        self._cache_scope: str = ""
         self._top_cache_time: float = 0.0
         self._top_cached_paths: list[str] = []
         self._fragment_hint: str | None = None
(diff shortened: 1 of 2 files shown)


hongquan:
Another approach: #1286

Issue #1480, opened 2026-03-17
kimi cli的 flow:skill，在交互模式下， 流程没法中断或者选择 || kimi cli's flow:skill, in interactive mode, the process cannot be interrupted or selected


Saro50:
### What version of Kimi Code CLI is running?

1.22.0

### Which open platform/subscription were you using?

/login

### Which model were you using?

kimi-for-coding

### What platform is your computer?

Darwin 25.3.0 arm64 arm

### What issue are you seeing?

使用示例的flow:skill
---
name: code-review
description: 代码审查工作流
type: flow
---

```d2
BEGIN -> B -> C
B: 分析现有代码，为 开发入口 编写说明文档
C: Review 设计文档是否足够详细
C -> B: 否
C -> D: 是
D: 开始实现
D -> END
```
kimi cli的 flow:skill，在交互模式下， 流程没法中断或者选择

### What steps can reproduce the bug?

任意项目，下  /flow:code-review  会一直输出得到最终结果，人为没法介入

---
### What version of Kimi Code CLI is running?

1.22.0

### Which open platform/subscription were you using?

/login

### Which model were you using?

kimi-for-coding

### What platform is your computer?

Darwin 25.3.0 arm64 arm

### What issue are you seeing?

Use example flow:skill
---
name: code-review
description: Code review workflow
type: flow
---

```d2
BEGIN -> B -> C
B: Analyze existing code and write documentation for the development portal
C: Review whether the design document is detailed enough
C -> B: No
C -> D: Yes
D: Start to realize
D->END
```
kimi cli's flow:skill, in interactive mode, the process cannot be interrupted or selected

### What steps can reproduce the bug?

For any project, /flow:code-review will always output the final result, and humans cannot intervene.

PR #1545, opened 2026-03-22
fix(keyboard): clear ISIG/IEXTEN in raw mode to prevent SIGTSTP suspension

### LINKED ISSUE #38: can not handle CTRL+Z singal

➜  kimi-cli git:(hy/support_ctrl_z) uv run kimi
╭─────────────────────────────────────────────────╮
│                                                 │
│   ▐█▛█▛█▌  Welcome to Kimi CLI!                 │
│   ▐█████▌  Send /help for help information.     │
│                                                 │
│  Directory: /Users/hyi/prs/kimi-cli             │
│  Session: d4b316ee-ff86-458f-91c8-de775612073c  │
│  Model: kimi-k2-turbo-preview                   │
│                                                 │
╰─────────────────────────────────────────────────╯
hyi✨ ^Z^Z^Z

### Additional information

_No response_

n-WN:
## Related Issue

Resolves #38 — `can not handle CTRL+Z singal` (reported by @yihong0618, closed as wontfix but never actually fixed).

Also related: #1480 (flow:skill not interruptible in interactive mode — the `_LiveView` path where this bug lives).

## Description

The raw keyboard listener (`keyboard.py`) used in the standalone `_LiveView` visualization path only cleared `ICANON` and `ECHO` when entering raw mode, but left `ISIG` and `IEXTEN` set. With `ISIG` active, the kernel intercepts control characters before they reach the application:

- `Ctrl+Z` → `SIGTSTP` → process suspended (no handler installed → hung process)
- `Ctrl+C` → `SIGINT` → process killed

This does **not** affect the normal interactive shell, which uses `prompt_toolkit` (it correctly clears `ISIG` via `tty.setraw`). The bug manifests in the `_LiveView` path — activated when `prompt_session` is `None` (e.g., piped/non-interactive runs, or when an IME interferes with tmux prefix detection causing raw keys to reach the app).

### Fix

**`keyboard.py`** (2 lines changed):
- Clear `ISIG` and `IEXTEN` alongside `ICANON` and `ECHO` when configuring raw mode, so all control characters are delivered as bytes instead of signals.
- Map `\x03` (Ctrl+C) to a new `KeyEvent.CTRL_C` enum value.

**`visualize.py`** (1 line changed):
- Handle `CTRL_C` the same as `ESCAPE` in `dispatch_keyboard_event` — set the cancel event to abort the active run.

### Before / After (verified in tmux)

| Metric | origin/main | This branch |
|--------|-------------|-------------|
| `ISIG` flag | **SET** | **CLEARED** |
| `IEXTEN` flag | SET | CLEARED |
| `Ctrl+Z` | SIGTSTP → process suspended | Byte `\x1a` → silently ignored |
| `Ctrl+C` | SIGINT → process killed | Byte `\x03` → `CTRL_C` event → cancels run |
| Arrow keys | ✅ | ✅ |

## Testing

- 18 new unit tests in `tests/ui_and_conv/test_keyboard_raw_mode.py` covering:
  - Raw mode flag verification (`ISIG`, `IEXTEN`, `ICANON`, `ECHO` all cleared)
  - All 14 byte → `KeyEvent` mappings (parametrized)
  - `CTRL_C` dispatches cancel event in `_LiveView`
  - `ESCAPE` still dispatches cancel (regression test)
  - `CTRL_C` is a no-op when no cancel event is set
- Full test suite: 1172 passed, 6 skipped, 0 failed
- Ruff lint + format: clean

## Checklist


yihong0618:
FYI, we should also consider the behavior on windows

n-WN:
## Bug Reproduction in Real kimi-cli (`--print` mode)

### Before (main): Ctrl+Z during `kimi --print` streaming
before
> [asciinema replay](https://asciinema.org/a/NXIBaebO1t0akZGB)

### After (this PR): Ctrl+Z silently consumed, Ctrl+C cancels gracefully
after
> [asciinema replay](https://asciinema.org/a/uVIHaXtWZwyfaBBG)

Both demos run `uv run kimi --yolo -p 'count slowly from 1 to 30'` which uses the `_LiveView` code path (the raw keyboard listener). Ctrl+Z is sent during active streaming.

PR #1817, opened 2026-04-09
fix(auth): retry with token refresh on 401 and harden OAuth resilience

n-WN:
## Related Issue

Fixes frequent "Authorization failed, please check your login status" errors reported by multiple users who need to re-login 3-4 times per day. Root cause confirmed via 16-minute real-environment reproduction experiment.

## Root Cause Analysis

Access tokens expire after **15 minutes** (`expires_in=900`), but the retry logic (written in Sep 2025 for static API keys) never included 401 in the retryable set. When OAuth was grafted onto the codebase in Jan 2026, only **proactive** background refresh was added — no **reactive** 401-then-refresh-then-retry path. This fails under:

- Laptop sleep >15 minutes (event loop frozen, token expires)
- Network instability during the narrow 5-minute refresh window
- Long tool-call chains (>10 min) where token expires mid-turn
- Multiple kimi-cli instances racing on credential file writes

Confirmed via real token experiment on macOS: token expired at T+15:04, API returned 401 within 5 seconds. Manual `refresh_token()` + retry immediately recovered to 200.

### Core fix: 401 reactive recovery (`kimisoul.py`)

`_run_with_connection_recovery()` now catches `APIStatusError(401)` for OAuth-enabled providers, calls `ensure_fresh(force=True)`, and retries the operation exactly once. Non-OAuth providers (plain API key) skip the retry to avoid useless latency.

### Token refresh hardening (`oauth.py`)

| Change | What | Why |
|--------|------|-----|
| `force=True` parameter | `ensure_fresh()` / `_refresh_tokens()` bypass expiry-threshold check | After 401, local `expires_at` may not have passed but server rejected the token |
| Dynamic refresh threshold | `max(300s, expires_in * 0.5)` instead of hardcoded 300s | 15-min tokens now get 7.5-min refresh window (7+ attempts vs 5) |
| Retry with backoff | `refresh_token()` retries 3x on network errors (1s, 2s, 4s) | Single network hiccup no longer burns the entire refresh window |
| Sleep/wake detection | Wall-clock elapsed check in refresh loop; force refresh if >2x interval | Catches laptop lid-close scenarios |
| `expires_in` on OAuthToken | Stored and persisted for proportional threshold | Backward-compatible (defaults to 0 for old tokens) |

### Credential file safety (`oauth.py`)

| Change | What | Why |
|--------|------|-----|
| Atomic writes | `tempfile.mkstemp()` + `os.replace()` | Prevents corrupt JSON from concurrent reads during write |
| Short-write guard | Check `os.write()` return value | Disk-pressure partial writes don't produce truncated JSON |
| chmod non-fatal | `suppress(OSError)` around `os.chmod` | Mounted volumes / restricted filesystems don't break token persistence |
| TOCTOU protection | `_load_from_file` catches `OSError` alongside `JSONDecodeError` | File deleted between `exists()` and `read_text()` returns None gracefully |

### Cross-process coordination (`oauth.py`)

| Change | What | Why |
|--------|------|-----|
| `_CrossProcessLock` | `fcntl.flock` (Unix) / `msvcrt.locking` (Windows) | Multiple kimi-cli instances (terminal + VS Code) coordinate refresh |
| Triple-check pattern | Check before lock -> acquire lock -> check again after lock | If another process refreshed while waiting, use its token |
| 5x retry + jitter | `asyncio.sleep(1 + random())` between lock attempts | Break thundering herd without blocking event loop |
| Graceful degradation | Lock failure logs warning, refresh proceeds uncoordinated | Lock is advisory — never blocks auth entirely |
| Windows fixes | Write `\0` sentinel for `msvcrt.locking`; explicit `LK_UNLCK` before close | `msvcrt` differs from `fcntl` in requiring pre-positioned bytes and explicit unlock |

### Token revocation handling (`oauth.py`)

| Change | What | Why |
|--------|------|-----|
| `delete_tokens(ref)` on revocation | Remove credential file when refresh token is truly rejected | Prevents stale token re-load on next cycle |
| 1-second grace window | Sleep before deletion to let concurrent instances persist rotated tokens | Avoids deleting another instance's freshly-written token |
| `if token.access_token:` guard | `ensure_fresh` won't apply empty access_token to runtime | Defense-in-depth against stale credential files |

### Tests (`tests/auth/test_oauth_refresh.py`)

9 new unit tests covering:
- Retry on transient network errors (success on 3rd attempt)
- No retry on `OAuthUnauthorized` (truly rejected)
- `OAuthError` raised after all retries exhausted
- Dynamic threshold triggers refresh at 50% lifetime
- Dynamic threshold skips when plenty of time remains
- `force=True` bypasses threshold
- Atomic file write (no temp files left, 0o600 permissions)
- `expires_in` roundtrip through save/load
- Backward compatibility with old token format (no `expires_in`)

## Before / After

| Scenario | Before | After |
|----------|--------|-------|
| Laptop sleep >15 min | 401 -> "Authorization failed" -> must `/login` | 401 -> auto-refresh -> retry -> seamless |
| Network hiccup in refresh window | 5 chances, no retry on failure -> token expires | 7+ chances, 3x retry with backoff per attempt |
| Two instances refresh simultaneously | No coordination -> possible token file corruption | File lock + triple-check -> only one refreshes |
| Token truly revoked | Token file deleted (could race) | Delete under lock + grace window + empty-token guard |
| Non-OAuth provider gets 401 | Would still attempt useless refresh | Skips retry, surfaces error immediately |

## Reproduction

Confirmed with a 16-minute real-token experiment (`/tmp/kimi-auth-lab/`):
```
T+00:00  Token issued (expires_in=900)
T+10:03  Entered refresh window (remaining < 5 min) — intentionally not refreshing
T+15:04  API returned 401 (token expired 5 seconds ago)
T+16:05  Manual refresh + retry -> 200 OK (fix validated)
```

## Note on PR scope

This PR addresses the complete OAuth resilience stack. If preferred, it can be split into:
- **PR A** (core fix): 401 retry in `kimisoul.py` + `refresh_token()` retry with backoff (~50 LOC)
- **PR B** (hardening): Atomic writes, dynamic threshold, `expires_in` persistence
- **PR C** (coordination): Cross-process file lock

Happy to split if that helps review.

## Checklist

Automated tests pass (2200 existing + 9 new). `make gen-changelog` and `make gen-docs` are pending — will run before final merge.

### DIFF
--- a/src/kimi_cli/auth/oauth.py
+++ b/src/kimi_cli/auth/oauth.py
@@ -4,6 +4,7 @@ import asyncio
 import json
 import os
 import platform
+import random
 import socket
 import sys
 import tempfile
@@ -53,6 +54,8 @@ KEYRING_SERVICE = "kimi-code"
 REFRESH_INTERVAL_SECONDS = 60
 MIN_REFRESH_THRESHOLD_SECONDS = 300
 REFRESH_THRESHOLD_RATIO = 0.5
+_CROSS_PROCESS_LOCK_TIMEOUT = 10  # seconds
+_CROSS_PROCESS_LOCK_RETRIES = 5
 
 
 class OAuthError(RuntimeError):
@@ -230,6 +233,62 @@ def _credentials_dir() -> Path:
 def _credentials_path(key: str) -> Path:
     name = key.removeprefix("oauth/").split("/")[-1] or key
     return _credentials_dir() / f"{name}.json"
+
+
+def _credentials_lock_path(key: str) -> Path:
+    name = key.removeprefix("oauth/").split("/")[-1] or key
+    return _credentials_dir() / f"{name}.lock"
+
+
+class _CrossProcessLock:
+    """File-based lock that coordinates token refresh across kimi-cli processes.
+
+    Uses fcntl.flock on Unix and msvcrt.locking on Windows.
+    """
+
+    def __init__(self, key: str) -> None:
+        self._path = _credentials_lock_path(key)
+        self._fd: int | None = None
+
+    def _acquire(self) -> bool:
+        try:
+            self._fd = os.open(str(self._path), os.O_CREAT | os.O_RDWR)
+        except OSError:
+            return False
+        try:
+            if sys.platform == "win32":
+                import msvcrt
+
+                msvcrt.locking(self._fd, msvcrt.LK_NBLCK, 1)
+            else:
+                import fcntl
+
+                fcntl.flock(self._fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            return True
+        except OSError:
+            os.close(self._fd)
+            self._fd = None
+            return False
+
+    def release(self) -> None:
+        if self._fd is not None:
+            with suppress(OSError):
+                os.close(self._fd)
+            self._fd = None
+
+    async def acquire_with_retry(self) -> bool:
+        for _attempt in range(_CROSS_PROCESS_LOCK_RETRIES):
+            if self._acquire():
+                return True
+            await asyncio.sleep(1 + random.random())
+            # After waiting, re-check if the token was refreshed by the holder.
+        return self._acquire()
+
+    async def __aenter__(self) -> bool:
+        return await self.acquire_with_retry()
+
+    async def __aexit__(self, *args: object) -> None:
+        self.release()
 
 
 def _load_from_keyring(key: str) -> OAuthToken | None:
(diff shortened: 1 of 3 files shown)


n-WN:
Superseded by split PRs for easier review:
- #1819 — Core fix: 401 retry with token refresh (~260 LOC, ship first)
- #1820 — Hardening: dynamic threshold, atomic writes, sleep/wake detection, cross-process lock (~530 LOC, builds on #1819)

PR #1819, opened 2026-04-09
fix(auth): retry with token refresh on 401 to prevent forced re-login

n-WN:
## Related Issue

Fixes frequent "Authorization failed" errors that force users to re-login 3-4 times per day. Confirmed via 16-minute real-token experiment: access tokens expire after 15 minutes, but the code had no reactive recovery path.

## Root Cause

The retry logic (`_is_retryable_error`, written Sep 2025 for static API keys) never included 401. When OAuth was added in Jan 2026, only proactive background refresh was implemented — no reactive 401→refresh→retry. This fails on laptop sleep >15 min, network hiccups, or long tool-call chains.

## Changes (minimal, ~260 LOC)

**`src/kimi_cli/auth/oauth.py`:**
- `ensure_fresh(force=True)` / `_refresh_tokens(force=True)` — bypass expiry-threshold check after 401 (server already rejected the token)
- `refresh_token()` retries 3x with exponential backoff (1s, 2s, 4s) on transient network errors; `OAuthUnauthorized` is never retried
- `ensure_fresh` guards `_apply_access_token` with `if token.access_token:` to prevent empty token overwrite

**`src/kimi_cli/soul/kimisoul.py`:**
- `_run_with_connection_recovery()` catches 401, checks active provider has OAuth, calls `ensure_fresh(force=True)`, retries once
- Non-OAuth providers skip retry (no useless latency)
- `raise error from refresh_exc` preserves both the 401 and refresh failure for diagnostics

**`tests/auth/test_oauth_refresh.py`** — 4 new tests:
- Retry succeeds on 3rd attempt after transient errors
- `OAuthUnauthorized` raises immediately (no retry)
- `OAuthError` after all retries exhausted
- `force=True` bypasses threshold check

## Follow-up PRs

This is intentionally minimal. Planned follow-ups:
- **Token lifecycle hardening**: dynamic refresh threshold, atomic credential writes, sleep/wake detection, revocation cleanup
- **Cross-process lock**: multi-instance coordination via fcntl/msvcrt

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoonshotAI/kimi-cli/blob/main/CONTRIBUTING.md) document.
- [x] I have linked the related issue, if any.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have run `make gen-changelog` to update the changelog.
- [ ] I have run `make gen-docs` to update the user documentation.

---

[...]

PR #1821, opened 2026-04-09
fix(auth): harden token lifecycle with dynamic threshold, atomic writes, and revocation cleanup

n-WN:
## Related Issue

Follow-up to #1819. Hardens the OAuth token lifecycle against edge cases discovered during investigation.

## Description

Builds on the 401 retry fix (#1819) with additional resilience. Each change addresses a specific failure mode confirmed via real-token experiments and AI code review.

### Dynamic refresh threshold
- `max(300s, expires_in × 0.5)` instead of hardcoded 300s
- 15-min tokens now get 7.5-min refresh window (7+ attempts vs 5)
- `expires_in` stored on `OAuthToken` (backward-compatible, defaults to 0)

### Atomic credential writes
- `_save_to_file` uses `tempfile.mkstemp()` + `os.replace()` (POSIX atomic)
- Short-write guard: checks `os.write()` return value
- `chmod(0o600)` wrapped in `suppress(OSError)` for restricted filesystems

### TOCTOU protection
- `_load_from_file` catches `OSError` alongside `JSONDecodeError`

### Sleep/wake detection
- Background refresh loop checks wall-clock elapsed time
- Forces immediate refresh when `elapsed > 2× interval`

### Token revocation cleanup
- `delete_tokens(ref)` on truly revoked tokens with 1s grace window
- `if token.access_token:` guard prevents empty token from overwriting runtime key

### Tests (5 new, 9 total with #1819)
- Dynamic threshold triggers/skips at correct boundaries
- Atomic write: correct file, no .tmp leftovers, 0o600 permissions
- `expires_in` roundtrip and backward compatibility

## Checklist

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/1821" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/kimi_cli/auth/oauth.py
+++ b/src/kimi_cli/auth/oauth.py
@@ -51,6 +51,7 @@ DEFAULT_OAUTH_HOST = "https://auth.kimi.com"
 KEYRING_SERVICE = "kimi-code"
 REFRESH_INTERVAL_SECONDS = 60
 REFRESH_THRESHOLD_SECONDS = 300
+_REFRESH_MAX_RETRIES = 3
 
 
 class OAuthError(RuntimeError):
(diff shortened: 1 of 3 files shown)


n-WN:
Superseded by #1822, which includes both commits from this PR plus cross-process lock, force propagation, and additional tests.

PR #1771, opened 2026-04-06
fix: always stringify tool message content in Chat Completions provider

he-yufeng:
## Summary

Fixes #1762

The OpenAI Chat Completions API requires `content` to be a **string** for `role: "tool"` messages. When a tool result had multiple `ContentPart`s (e.g. a system-reminder TextPart + the actual output TextPart), `_convert_message` kept them as an array, causing:

```
400: Failed to deserialize the JSON body into the target type:
messages[3]: invalid type: sequence, expected a string
```

The fix: always collapse tool message content to a single `TextPart` via `extract_text()`, regardless of the `tool_message_conversion` setting. This matches the API spec and is consistent with how single-part tool results already worked.

## Changes

- `openai_legacy.py`: Remove the `tool_message_conversion == "extract_text"` condition for tool messages - they always need string content
- `test_openai_legacy.py`: Update snapshots to reflect string content for multi-part tool results

## Test plan

- [ ] `tool_call_with_image` snapshot: content is now `"5"` (string) instead of `[{text: "5"}, {image_url: ...}]` (array)
- [ ] `parallel_tool_calls` snapshot: content is now collapsed string instead of array of TextParts
- [ ] Existing tests still pass

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/1771" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

Issue #1344, opened 2026-03-05
API Error: 400 {"error":{"type":"invalid_request_error","message":"Invalid request Error"},"type":"error"}


Lyken17:
### What version of Kimi Code CLI is running?

kimi, version 1.14.0

### Which open platform/subscription were you using?

Kimi Code

### Which model were you using?

kimi-for-coding

### What platform is your computer?

Darwin 25.2.0 arm64 arm

### What issue are you seeing?

QQ

### What steps can reproduce the bug?

with `kimi-cli`, current every works smoothly, but when switching to claude code (endpoint set to kimi), it only works for daily chats. When `toolcall` are involved, it returns API-400 err

```
➜  v2-dev git:(main) claude -p "how are you"
I'm doing well, thanks for asking. Ready to help you with whatever you're working on. How can I assist you today?
➜  v2-dev git:(main) claude -p "list the files"
API Error: 400 {"error":{"type":"invalid_request_error","message":"Invalid request Error"},"type":"error"}
```

Lyken17:
similar issue has also been observed in https://github.com/MoonshotAI/kimi-cli/issues/778

I tested opencode and it works as usual, so this should be specific bug related to claude and https://api.kimi.com/coding/

jichaofdu:
I have the same problem when using cluade code
 It occurs from this morning.

ShiliangLuo:
I restart vscode for its aotoupdate, and then i have same problem.

quansang:
+1

xraywu:
the error occurs as soon as a SKILL is loaded, seemed to be an issue with Claude Code's agent ability.

carl10086:
I suspect the CC client switched to the latest model, which is causing the 400 error. When I use ⁠/model to switch back to the older 4.5 version, it seems to work fine.

```bash
  /model to try Opus 4.6

❯ /model
  ⎿  Set model to Default (claude-sonnet-4-6)

❯ /gc
  ⎿  API Error: 400 {"error":{"type":"invalid_request_error","message":"Invalid request Error"},"type":"error"}

❯ /model
  ⎿  Set model to haiku (claude-haiku-4-5-20251001)

❯ /gc

⏺ Bash(git status)
  ⎿  On branch main
     Your branch is up to date with 'origin/main'.
     … +17 lines (ctrl+o to expand)

⏺ Bash(git diff --stat HEAD)
  ⎿   pyproject.toml     |  1 +
[...]

Issue #778, opened 2026-01-29
【bug] API Error: 400 {"error":{"type":"invalid_request_error","message":"Invalid request Error"},"type":"error"}


DayDreammy:
### What version of Kimi Code CLI is running?

 Version: 2.1.23

### Which open platform/subscription were you using?

win11 , powershell

### Which model were you using?

 Model: Default (claude-sonnet-4-5-20250929)

### What platform is your computer?

Microsoft Windows NT 10.0.26100.0 x64

### What issue are you seeing?

<img width="2769" height="867" alt="Image" src="https://github.com/user-attachments/assets/fcde93e5-0aed-4655-b004-be561488c9b1" />

### What steps can reproduce the bug?

仔细阅读 "E:\问kimi-4个文件\Task.md"，结合文件内容，深度思考，规划实施方案，然后和我确认。

● Read 1 file (ctrl+o to expand)

● 我已经阅读了任务文件。这是一个深度个人健康数据分析项目，需要综合四个维度的数据进行跨学科分析。让我先探索数据文件的目录结构，了解具体的文件格式和
  容。

● Searched for 2 patterns, read 1 file (ctrl+o to expand)

● 找到了数据文件。让我读取这些文件以了解数据格式。

● Read 4 files (ctrl+o to expand)
  ⎿  API Error: 400 {"error":{"type":"invalid_request_error","message":"Invalid request Error"},"type":"error"}

✻ Churned for 1m 10s

❯ continue
  ⎿  API Error: 400 {"error":{"type":"invalid_request_error","message":"Invalid request Error"},"type":"error"}

2026.1.27 13点之前重试了两次同样的情况，17点第三次尝试暂时没有发现问题。

bkerf:
问题依旧! 这问题导致使用claude code时根本无法使用!

---
The problem remains! This problem makes it impossible to use claude code at all!

JMYok:
> 问题依旧! 这问题导致使用claude code时根本无法使用!
> 
> The problem remains! This problem makes it impossible to use claude code at all!

# 我做的尝试：
1.换API key（问题依然存在）
2.换CLI：使用opencode。问题就解决了。

xraywu:
只要一加载skills或者复杂的内容就会出现

murphypei:
真的很垃圾，接入 cc switch 也一直报这个错误。我还以为是我哪里没搞好，浪费我半天配置和查资料的时间。之前测试GLM完全没遇到过类似的问题。
[...]

PR #1978, opened 2026-04-21
fix(kosong/anthropic): merge parallel tool_results into one user msg

### LINKED ISSUE #1975: kosong Anthropic provider splits parallel tool_result blocks into multiple user messages — violates Anthropic Messages API spec

kimi-cli 1.37.0 (kosong 0.50.0), repro is code-level, reproducible on `main` at commit `f8f0ade8`.

N/A — this is a structural bug in `kosong`'s Anthropic chat provider, independent of subscription. Surfaces on any backend that implements Anthropic Messages API strictly (e.g. DeepSeek `/anthropic`).

Any model routed through `type = "anthropic"` provider. Originally observed with DeepSeek Anthropic-compatible endpoint.

Darwin 25.2.0 arm64 arm (platform-independent; the bug is in the wire-format conversion).

Per the [Anthropic Messages API spec](https://docs.claude.com/en/api/messages), when the assistant emits multiple `tool_use` blocks in a single turn (parallel tool calls), **all corresponding `tool_result` blocks must be packed into a single `user` message**:

```jsonc
{ "role": "assistant", "content": [tool_use_A, tool_use_B] }
{ "role": "user",      "content": [tool_result_A, tool_result_B] }   // one user, two blocks
```

Kosong's Anthropic provider instead emits one `user` message per `tool_result`:

```jsonc
{ "role": "assistant", "content": [tool_use_A, tool_use_B] }
{ "role": "user",      "content": [tool_result_A] }   // split
{ "role": "user",      "content": [tool_result_B] }
```

This has **two distinct impacts**, not one:

1. **Strict Anthropic-compatible backends hard-fail.** DeepSeek's `/anthropic` endpoint returns HTTP 400 on the second consecutive `user` message. Visible failure.
2. **Anthropic's own backend silently degrades.** `api.anthropic.com` leniently merges consecutive same-role turns so requests don't 400 — but the official parallel-tool-use docs ([troubleshooting section](https://platform.claude.com/docs/en/agents-and-tools/tool-use/parallel-tool-use#troubleshooting)) explicitly warn that the split form *"teaches"* Claude to avoid parallel tool calls in subsequent turns. Invisible degradation affecting every user doing multi-turn agent work, not just DeepSeek users.

Because impact #2 is silent, this bug likely went unnoticed for a long time despite being in the critical path of parallel-tool-use behavior.

**Root cause (two code locations in `packages/kosong/src/kosong/contrib/chat_provider/anthropic.py` on `main`):**

- `_convert_message`, `elif role == "tool"` branch (~L440–448 on `main`):
  ```python
  elif role == "tool":
      ...
      block = _tool_result_message_to_block(message.tool_call_id, content)
      return MessageParam(role="user", content=[block])   # each tool message → its own user message
  ```
- `generate`, history-to-wire loop (~L298–300 on `main`):
  ```python
  messages: list[MessageParam] = []
  for message in history:
      messages.append(self._convert_message(message))   # per-message append, no merge
  ```

Together these produce N separate `user` messages when the internal history contains N consecutive `role="tool"` messages from a parallel assistant turn.

**Reproducer (no server call required):** the existing snapshot test already encodes the buggy shape, so it both documents the current behavior and fails any attempted fix that doesn't update the snapshot.

```
uv run pytest packages/kosong/tests/api_snapshot_tests/test_anthropic.py -k parallel_tool_calls -v
```

The snapshot at `packages/kosong/tests/api_snapshot_tests/test_anthropic.py` (the `parallel_tool_calls` case) currently contains:

```jsonc
{ "role": "user", "content": [{"type": "tool_result", "tool_use_id": "call_add", ...}] },
{ "role": "user", "content": [{"type": "tool_result", "tool_use_id": "call_mul", ...}] }
```

which is the split/non-spec shape. It passes because code + snapshot are "wrong together."

`generate()` should merge consecutive `role="user"` messages whose contents are entirely `tool_result` blocks into a single `user` message before sending to the wire, so the request matches the Anthropic spec:

```jsonc
{ "role": "user", "content": [tool_result_A, tool_result_B] }
```

The `parallel_tool_calls` snapshot should be updated to the merged shape as part of the fix.

### Additional information

**Related issues (for triage cross-reference):**

- #1344, #778 — mirror direction: external Claude Code → Kimi `/anthropic` endpoint returns HTTP 400 on tool calls. Likely the same class of strict-vs-lenient parallel-tool-result handling, just on the inbound server side; worth investigating alongside this fix to make sure both ends are spec-aligned.
- #382 — prior CLOSED bug where DeepSeek's `/anthropic` edge behavior (null `usage.input_*`) broke kosong. Same pattern: kosong was too lax, strict compat backend exposed it.
- #1762 / PR #1771 — structurally identical fix in the OpenAI Chat Completions provider (tool message `content` shape had to be normalized to match spec rather than relying on backend lenience). Same thinking should apply here.

**Scope note:** This is strictly a provider-layer (wire-format) bug. No internal `Message` model changes are needed — the fix is local to `anthropic.py::generate()`.

wbxl2000:
## Related Issue

Resolve #1975

### Fix

`generate()` in the Anthropic chat provider now merges consecutive user messages whose contents are entirely `tool_result` blocks into a single user message before sending on the wire, matching the Anthropic Messages API [parallel-tool-use spec](https://platform.claude.com/docs/en/agents-and-tools/tool-use/parallel-tool-use#troubleshooting).

Also updates the `parallel_tool_calls` snapshot in `test_anthropic.py` (which had locked in the buggy split shape), and adds an assertion-style regression test that documents the spec-required shape independently of the snapshot — so if anyone regenerates the snapshot against buggy output again, the assertion test still fails.

### Why this matters beyond DeepSeek

Two distinct impacts, as documented in #1975:

1. **Strict Anthropic-compatible backends hard-fail.** DeepSeek's `/anthropic` endpoint returns HTTP 400 on the second consecutive user message. Visible failure.
2. **Anthropic's own backend silently degrades.** `api.anthropic.com` leniently merges consecutive same-role turns so requests don't 400 — but Anthropic's parallel-tool-use docs explicitly warn the split form *"teaches"* Claude to avoid parallel tool calls in subsequent turns. Invisible degradation affecting every multi-turn agent session, not just DeepSeek users.

Because impact #2 is silent, this has likely been quietly reducing parallel-tool-use across all Anthropic-backed models in kimi-cli.

### Scope

Provider-layer wire-format fix only. No internal `Message` model changes; other providers (OpenAI, Google, Kimi) are untouched. Fix is ~20 lines in `anthropic.py::generate()` plus one private helper `_is_tool_result_only`, following the existing `_is_*` / `cast()` patterns in the same file.

### Related issues

- #1344, #778 — mirror direction (Claude Code → Kimi `/anthropic` 400s on tool calls). Likely the same class of bug on the inbound server side; worth a separate audit.
- #1762 / #1771 — structurally identical prior fix in the OpenAI Chat Completions provider.
- #382 — prior CLOSED DeepSeek `/anthropic` edge case (null `usage.input_*`). Same pattern: kosong too lax, strict compat backend exposed it.

### Test plan

- New regression test `test_anthropic_parallel_tool_results_merged_into_single_user_message` — asserts 3 messages total, both `tool_result` blocks in the trailing user message, correct `tool_use_id`s.
- Updated `parallel_tool_calls` snapshot matches the merged shape; `cache_control` correctly lands on the last `tool_result` block after merging (same block as before, now just colocated with its sibling).
- Edge-case verification of the merge loop:
  - Single (non-parallel) tool call — no merge triggered.
  - 3 parallel tool calls — collapse to one user message with 3 `tool_result` blocks.
  - Parallel tool results followed by a user text message — only the tool results merge; text stays in its own message.
- `ruff check`, `ruff format --check`, `pyright` — all clean on modified files.

### Note on pre-existing env issue

On `main`, 20 tests in `tests/api_snapshot_tests/test_anthropic.py` are failing locally with `APIConnectionError` due to a `respx` ↔ `AsyncAnthropic` httpx-transport compatibility issue. This is orthogonal to this PR; the new regression test uses the same `respx` pattern as existing tests and will recover together with them once the env issue is fixed.

## Checklist

- [x] I have run `make gen-docs` to update the user documentation. (No user-facing docs changed; this is a provider-layer wire-format fix. Confirmed with maintainer that doc updates are not needed.)

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/1978" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Only write entries that are worth mentioning to users.
 
 ## Unreleased
 
+- Kosong: Fix parallel tool results being split into multiple user messages in Anthropic provider — consecutive tool-result-only user messages are now merged into a single message, complying with the Anthropic Messages API spec that all `tool_use` blocks in an assistant turn must be answered within one user message; this fixes 400 errors on strict Anthropic-compatible backends (e.g. DeepSeek `/anthropic` endpoint) and prevents the official backend from silently teaching the model to avoid parallel tool calls
+
 ## 1.37.0 (2026-04-20)
 
 - Print: Wait for background tasks before exiting — in one-shot `--print` mode, the process now waits for running background agents to finish and lets the model process their results, instead of exiting and killing them. The wait is capped at `min(max(active_task.timeout_s or agent_task_timeout_s), print_wait_ceiling_s)` (default ceiling 1h); on timeout the tasks are killed and the model gets one more turn via a `<system-reminder>` to summarise before exit
--- a/docs/en/release-notes/changelog.md
+++ b/docs/en/release-notes/changelog.md
@@ -4,6 +4,8 @@ This page documents the changes in each Kimi Code CLI release.
 
 ## Unreleased
 
+- Kosong: Fix parallel tool results being split into multiple user messages in Anthropic provider — consecutive tool-result-only user messages are now merged into a single message, complying with the Anthropic Messages API spec that all `tool_use` blocks in an assistant turn must be answered within one user message; this fixes 400 errors on strict Anthropic-compatible backends (e.g. DeepSeek `/anthropic` endpoint) and prevents the official backend from silently teaching the model to avoid parallel tool calls
+
 ## 1.37.0 (2026-04-20)
 
 - Print: Wait for background tasks before exiting — in one-shot `--print` mode, the process now waits for running background agents to finish and lets the model process their results, instead of exiting and killing them. The wait is capped at `min(max(active_task.timeout_s or agent_task_timeout_s), print_wait_ceiling_s)` (default ceiling 1h); on timeout the tasks are killed and the model gets one more turn via a `<system-reminder>` to summarise before exit
--- a/docs/zh/release-notes/changelog.md
+++ b/docs/zh/release-notes/changelog.md
@@ -4,6 +4,8 @@
 
 ## 未发布
 
+- Kosong：修复 Anthropic 供应商将并行工具结果拆分到多个 user message 的问题——现在会将仅包含工具结果的连续 user message 合并为单条消息，以符合 Anthropic Messages API 规范（assistant 一轮中的所有 `tool_use` 必须在同一条 user message 内回答）；修复了严格兼容后端（如 DeepSeek `/anthropic` 接口）返回 400 错误的问题，并避免官方后端静默地引导模型放弃并行工具调用
+
 ## 1.37.0 (2026-04-20)
 
 - Print：退出前等待后台任务完成——在单次 `--print` 模式下，进程现在会等待仍在运行的后台 Agent 完成并让模型处理它们的结果，而不是直接退出并杀死它们。等待时长上限为 `min(max(active_task.timeout_s or agent_task_timeout_s), print_wait_ceiling_s)`（默认上限 1 小时）；超时后杀死任务并通过 `<system-reminder>` 给模型最后一轮机会向用户总结后再退出
--- a/packages/kosong/CHANGELOG.md
+++ b/packages/kosong/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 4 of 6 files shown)


wbxl2000:
**DeepSeek 端点验证**

失败的情况:

before

成功的情况:

Snipaste_2026-04-22_13-56-26

**端到端测试**

失败的情况:

e2e-before

修复后:

e2e-success

Issue #1940, opened 2026-04-19
Count auth failure problem when using kimi code in vscode or cursor.


Eglusaxie:
### What version of Kimi Code CLI is running?

idk, I'm using it through vscode and cursor,newst version in marketplace v0.5.3

### Which open platform/subscription were you using?

vscode and cursor,both of them.

### Which model were you using?

kimi-for-coding

### What platform is your computer?

windows

### What issue are you seeing?

Frequently count red alert when my agent session runing .The message is like:
```
Authentication failed. Your login session may have expired. Please run "/login" to sign in again.
{"jsonrpc":"2.0","id":"2_177660*******","error":{"code":-32004,"message":"Authentication failed. Your login session may have expired. Please run \"/login\" to sign in again.","data":null}}
```
tried type /login in the GUI of plugin of vscode, but no reponse.
And the red alert massage come with a retry button, click it will continuing session for a while until next same alert recur.

### What steps can reproduce the bug?

normly install the plugin, login, create session.

Issue #1547, opened 2026-03-22
Repeated error for "Authorization failed, please check your login status" mid generation


Philipp-Pfeiffer:
### What version of Kimi Code CLI is running?

kimi, version 1.24.0

### Which open platform/subscription were you using?

Kimi Code

### Which model were you using?

kimi-for-coding

### What platform is your computer?

Linux 6.12.73-1-lts x86_64 unknown 

### What issue are you seeing?

"Normal Output with random cutoff"
Authorization failed, please check your login status

<img width="584" height="64" alt="Image" src="https://github.com/user-attachments/assets/c89e7da0-d3ac-4b42-aaa8-9c9d0c771a73" />

Doesn't happen on specific prompts but often.

### What steps can reproduce the bug?

Happens under normal usage, interrupts pretty often. No specific reproduction

### What is the expected behavior?

Expected behavior is to not ask me to login every other prompt.

Issue #1350, opened 2026-03-05
频繁出现 Authorization failed, please check your login status


dapeng1162:
### What version of Kimi Code CLI is running?

kimi, version 1.17.0

### Which open platform/subscription were you using?

/login

### Which model were you using?

kimi-for-coding

### What platform is your computer?

debian12

### What issue are you seeing?

频繁的Authorization failed, please check your login status

### What steps can reproduce the bug?

用着用着就会出现

PR #1996, opened 2026-04-22
fix(auth): do not delete credentials file on a single refresh 401

wbxl2000:
## Related Issue

Refs #1547, #1350, #1940. Addresses the frequent "Authorization failed. Your session may have expired." symptom that forces OAuth users to `/login` repeatedly during normal use.

## The bug in one sentence

`delete_tokens(ref)` removes **whatever the ref currently points to on disk**, not "the refresh_token that just got 401". In single-instance usage those are the same file content; in multi-instance usage they aren't — a concurrent `OAuthManager` may have legitimately rotated and written a valid new token into the same file between the `load_tokens` check at `src/kimi_cli/auth/oauth.py:954` and the `delete_tokens` call at `:964`. We wipe it.

## Why multi-instance races are common here

`OAuthManager` instances aren't a per-process singleton; several places in the codebase instantiate their own, sharing `~/.kimi/credentials/kimi-code.json`:

- `src/kimi_cli/app.py:199` — main entry
- `src/kimi_cli/web/api/sessions.py:817` — **per session title-generation** (added in #1658)
- `src/kimi_cli/cli/plugin.py:212`, `src/kimi_cli/plugin/tool.py:33` — plugin paths
- ACP worker subprocess (independent process)

Each has its own `_refresh_lock` (asyncio, scoped to the instance). Cross-instance coordination relies only on the file lock (`_CrossProcessLock`), which itself has a degraded-to-no-lock fallback (`oauth.py:946`). Access tokens last only 15 min (confirmed by #1819's real-token experiment), so refresh fires every ~7.5 min per manager — plenty of race opportunities.

## Fix

**1) Drop `delete_tokens(ref)`.** Clearing the in-memory cache remains. Convergence is still correct:

- If the refresh_token is **truly revoked**: next `ensure_fresh` 401s again, user is prompted to `/login`, which atomically overwrites the file. Same UX as before, one extra network round-trip.
- If a **concurrent instance just rotated**: file now holds the valid new token; the next `resolve_api_key` picks it up, user keeps working.

**2) Add a process-local rejection tombstone** (`_REJECTED_REFRESH_TOKENS`). Removing `delete_tokens` alone would regress users who configure both OAuth AND a static `api_key` fallback: `resolve_api_key` prefers OAuth over api_key by design, so a stale-but-undeleted OAuth file would keep shadowing the api_key. The tombstone records the rejected refresh_token with a 5-minute cooldown, and while live:

- `resolve_api_key` does not reload the persisted token from disk → the static api_key fallback takes over.
- `_apply_access_token` falls back to the provider's configured api_key on the live LLM client → in-flight sessions recover without waiting for the next turn.
- Background `ensure_fresh` short-circuits and does not re-hit the server with the same dead refresh_token.
- The tombstone clears automatically when the on-disk refresh_token differs from the rejected one (i.e. a concurrent instance rotated, or `/login` overwrote the file), and is also explicitly cleared in every rotation-recovery branch.

The tombstone is module-level rather than per-manager — OAuth credentials are a process-wide resource, so all managers in this process should see the same rejection state.

## Scope

- `src/kimi_cli/auth/oauth.py`:
  - Remove `delete_tokens(ref)` in the `OAuthUnauthorized` handler of `_refresh_tokens` and rewrite the surrounding comment (ref-is-a-pointer-not-a-value framing).
  - Introduce `_REJECTED_REFRESH_TOKENS` module dict and four helpers on `OAuthManager`: `_rejected_refresh_state`, `_should_suppress_persisted_token`, `_can_retry_rejected_refresh_token`, `_mark_refresh_token_rejected`, `_clear_rejected_refresh_token`. Call sites in `ensure_fresh`, `_refresh_tokens`, `resolve_api_key`, and `_load_initial_tokens`.
  - `_apply_access_token` falls back to the provider's configured `api_key` when the OAuth access token is empty, so the live LLM client switches over immediately when OAuth is cleared.
- `tests/auth/test_oauth_refresh.py`:
  - Drop a now-redundant `patch("delete_tokens")` from `test_ensure_fresh_force_raises_on_unauthorized`.
  - Add `patch("asyncio.sleep")` to tests that exercise the `OAuthUnauthorized` handler so they no longer wait 1 s.
  - New `test_unauthorized_must_not_delete_credentials_file` (force=True path, real tmp fs) — fails on main, passes with this change.
  - New `test_unauthorized_non_force_must_not_delete_credentials_file` (background-refresh path).
  - New `test_rejected_refresh_token_cooldown_skips_background_retry` (cooldown is observed).
  - New `test_rejected_tombstone_cleared_when_concurrent_instance_rotated` (auto-heal on disk rotation).
  - Autouse fixture clears the module-level tombstone dict between tests.
- `tests/auth/test_resolve_api_key.py`:
  - Autouse fixture for tombstone isolation.
  - New `test_resolve_api_key_falls_back_after_rejected_refresh_token` — proves the api_key fallback path.
- Does not touch `_CrossProcessLock`, the defense layers upstream, or ACP / wire / web 401 handling.

## Why this is still only a partial fix (follow-up work)

The TOCTOU is a symptom of a deeper architectural issue. Suggested separate PRs:

- **`refactor(auth)`: single OAuthManager per process** — eliminates all in-process races; cross-process becomes the only remaining surface. High priority.
- **`fix(auth)`: make `_CrossProcessLock` not degrade silently** — remove the "5 retries then no-lock fallback" path. Medium.
- **`audit(auth)`: unify 401 handling across shell / acp / wire** — ACP (`session.py:221`) and wire (`server.py:668`) raise `auth_required` immediately instead of routing through `_run_with_connection_recovery`, which likely contributes to #1940's "retry button keeps popping up" loop. Medium.

The fix in this PR remains correct regardless of whether the above land — "a single 401 from the refresh endpoint is not sufficient evidence to destroy persistent credentials" is a permanent principle that no future refactor should reintroduce.

## Checklist

- [ ] I have run `make gen-docs` to update the user documentation. — **N/A**: internal behavior change, no user-facing docs affected.


wbxl2000:
@codex review

zouyee:
```
Interrupted
Authorization failed. Your session may have expired.
Type /login to re-authenticate.
Server: Error code: 401 - {'error': {'message': 'The API Key appears to be invalid or
may have expired. Please verify your credentials and try again.', 'type': 'invalid_aut
hentication_error'}}
```
It seems there is still a problem. I have just upgraded kimi-cli.

```
 kimi-cli --version
kimi, version 1.38.0
kimi --version
kimi, version 1.38.0
```

Issue #1663, opened 2026-03-30
LLM provider error 400 "text content is empty" when processing image input from MCP tools


xxYaOho:
## Description

When using MCP tools that return image data (e.g. Sketch MCP `get_selection_as_image`, `export`), the CLI frequently triggers a 400 error from the LLM provider, causing the conversation to freeze and become unresponsive.

## Error

```
LLM provider error: Error code: 400 - {'error': {'message': 'text content is empty', 'type': 'invalid_request_error'}}
```

## Reproduction

1. Use kimi-cli in a project with Sketch MCP configured
2. Ask the agent to capture a screenshot from Sketch, e.g. `sketch_get_selection_as_image` or `sketch.export(layer, { output: false }).toNSData()`
3. The MCP tool executes successfully and returns image data
4. When the response is sent to the LLM provider, it triggers the 400 error above
5. The CLI becomes unresponsive — you cannot continue the conversation

## Observations

- This seems to happen consistently when MCP tools return image/binary content
- The error suggests the message being sent to the LLM has an empty `text` field, likely because the response contains only image content without accompanying text
- After the error occurs, the CLI is stuck and the user has to restart the session

## Expected Behavior

- Image data returned from MCP tools should be properly formatted in the message payload sent to the LLM provider
- If the provider doesn't support image input, it should gracefully handle the error and allow the conversation to continue
- The CLI should not freeze after encountering this error

## Environment

- kimi-cli (latest)
- Sketch MCP integration
- macOS

## Workaround

Restart the CLI session after the error occurs.

PR #2035, opened 2026-04-23
fix(kosong/kimi): omit content field on empty assistant tool-call messages

wbxl2000:
## Related Issue

No pre-existing issue filed (reported directly by a user running into it).
Related: #1663 — same `400 "text content is empty"` error, different trigger
path (MCP tool returning image data). Not fixed by this PR.

### The bug

Kimi-for-Coding (`api.kimi.com/coding/v1`) rejects `chat.completions` requests
whose assistant message has `content` of shape
`[{"type": "text", "text": ""}]` with `400 "text content is empty"`.

This shape arises naturally in normal tool-call flow:

1. Turn 1: the model emits reasoning tokens + a tool call, no user-visible text.
2. kimi-cli stores the reply as `Message(content=[TextPart(text="")], tool_calls=[...])`.
3. Turn 2: `_convert_message` serializes this for replay. Before this fix, the
   wire body contained an empty text part, so the compat layer returned 400 and
   the entire conversation became unrecoverable.

### The fix

In `_convert_message`, when the assistant message has `tool_calls` AND the visible
content is effectively empty (all `TextPart`s are whitespace-only), drop the
`content` field from the outbound body. OpenAI-compatible APIs explicitly allow
omitting `content` on assistant tool-call messages, so this is strictly safer.

Unaffected paths (verified via snapshot tests):
- assistant messages with real text content
- assistant messages containing non-text parts (images, audio, video)
- `user` / `tool` / `system` messages

### Verification

- Snapshot test added to `packages/kosong/tests/api_snapshot_tests/test_kimi.py`
  asserting the wire body omits `content` in this case.
- E2E test added at `tests/e2e/test_kimi_empty_tool_call_content_e2e.py` using a
  strict mock compat server that returns 400 for effectively-empty content.
- Verified end-to-end against the real `api.kimi.com/coding/v1`:
  - after fix: 200 (request accepted)
  - injecting `content: [{"type":"text","text":""}]` back into the same body:
    `400 "text content is empty"` — reproduces the original error.

## Checklist

- [x] I have linked the related issue (#1663 as context; no pre-existing issue for this exact trigger).
- [x] I have added tests that prove my fix is effective.
- [ ] I have run `make gen-docs` to update the user documentation. **N/A** — this fix has no user-visible behavior change: no new/removed flag, no configuration change, no CLI output change. It only prevents a specific wire shape from being rejected by the provider.

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2035" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


wbxl2000:
@codex

PR #2046, opened 2026-04-23
fix(ui): flip /usage gauge color thresholds

### LINKED ISSUE #2019: Wrong usage color

1.37.0

It uses wrong color in the usage command.

The color is chosen by the remaining ratio:

https://github.com/MoonshotAI/kimi-cli/blob/34025f6af57c60de5b039673dfd607f18e2e903b/src/kimi_cli/ui/shell/usage.py#L255-L257
https://github.com/MoonshotAI/kimi-cli/blob/34025f6af57c60de5b039673dfd607f18e2e903b/src/kimi_cli/ui/shell/usage.py#L276-L281

```diff
+ # remaining ratio is passed in
  def _ratio_color(ratio: float) -> str:
-     if ratio >= 0.9:
+     if ratio <= 0.1:
          return "red"
-     if ratio >= 0.7:
+     if ratio <= 0.3:
        return "yellow"
     return "green"
```

he-yufeng:
Fixes #2019.

`_format_row` in `src/kimi_cli/ui/shell/usage.py` computes

```python
ratio = (row.limit - row.used) / row.limit if row.limit > 0 else 0
```

i.e. the **remaining** fraction (and the caller renders it as `{percent:.0f}% left`). The color helper below reads `ratio` as if it were a "used" gauge:

```python
def _ratio_color(ratio: float) -> str:
    if ratio >= 0.9:
        return "red"      # actually: 90%+ remaining, should be green
    if ratio >= 0.7:
        return "yellow"   # 70%+ remaining, should be green too
    return "green"        # < 70% remaining, gauge silently goes green as it drains
```

So the progress bar is green while usage is almost full and turns red right after a reset. This PR flips the thresholds so `red` kicks in when ≤10% remains and `yellow` when ≤30% remains, which matches the `% left` label.

Reporter (@sunmy2019) also gave the same fix in the issue. Added a short comment explaining the direction since the variable name alone doesn't make it obvious.

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2046" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/kimi_cli/ui/shell/usage.py
+++ b/src/kimi_cli/ui/shell/usage.py
@@ -274,8 +274,10 @@ def _format_row(row: UsageRow, label_width: int) -> RenderableType:
 
 
 def _ratio_color(ratio: float) -> str:
-    if ratio >= 0.9:
+    # ratio is the remaining fraction (1.0 = full, 0.0 = exhausted),
+    # so the thresholds are flipped from a "used" gauge.
+    if ratio <= 0.1:
         return "red"
-    if ratio >= 0.7:
+    if ratio <= 0.3:
         return "yellow"
     return "green"


wbxl2000:
Thanks for the clear fix and explanation here. The diagnosis was right: `/usage` was treating a remaining-quota ratio with used-quota color thresholds.

We have now merged #2078, which includes this threshold fix and also handles the related progress-fill direction, full-quota `finished_style`, unusual API values, and regression coverage. #2019 is closed by that merge, and #2078 credits this PR among the related fixes that helped converge on the final behavior.

Closing this as superseded by #2078 rather than because the contribution was wrong. Appreciated.

wbxl2000:
One more maintainer note: I really appreciate this PR. We ended up merging #2078 because it bundled this color-threshold fix with a few adjacent rendering edge cases and broader regression coverage, but your diagnosis here was correct and useful.

Please do not read the close as a rejection of the work. This was a duplicate-resolution call after the fix had landed, not a judgment that this PR was not valuable. Thanks again for helping move this issue forward.

PR #2115, opened 2026-04-29
fix(clipboard): enable clipboard paste on headless Linux over SSH

ZichenWen1:
## Problem

When running Kimi CLI on a headless Linux machine (e.g. a remote development server accessed via SSH), the `Ctrl+V` shortcut for pasting clipboard content is completely disabled. This happens because `pyperclip.paste()` fails during initialization when the `DISPLAY` environment variable is not set, causing `is_clipboard_available()` to return `False`.

However, the image paste pipeline in Kimi CLI does **not** actually depend on `pyperclip`. It can read images through `xclip` or `wl-paste` (e.g. via clipboard bridging tools like [cc-clip](https://github.com/ShunmeiCho/cc-clip) that shim `xclip` over an SSH tunnel).

The previous code used a single `is_clipboard_available()` boolean to gate two independent capabilities: (1) binding the `Ctrl+V` keybinding for media paste, and (2) creating a `PyperclipClipboard` for `PromptSession`. The latter is used by prompt_toolkit's built-in emacs keybindings (`ctrl-k`, `ctrl-w`, `ctrl-y`) without error handling, so a broken `PyperclipClipboard` would crash the UI.

Additionally, Pillow's `ImageGrab.grabclipboard()` on Linux has opaque internal tool selection: if both `wl-paste` and `xclip` are present, it may pick `wl-paste` first (even when it's broken), fail, and never fallback to `xclip`. This leaves xclip-bridge users unable to paste images in a common environment.

### 1. `src/kimi_cli/utils/clipboard.py`

- Added `import io` and `import subprocess`
- Restored `is_clipboard_available()` to a **strict pyperclip-only check** (as it was before).
- Added `is_media_clipboard_available()` that checks for `xclip` or `wl-paste` on Linux. On macOS and Windows it returns `True`.
- **New**: Added `_grab_image_linux()` that replaces Pillow's opaque `ImageGrab.grabclipboard()` Linux path with explicit xclip/wl-paste calls, ensuring xclip-bridge users are not blocked by a broken `wl-paste`.
  - **Session-aware ordering**: Uses `WAYLAND_DISPLAY`/`DISPLAY` env vars to pick the matching backend first (Wayland → wl-paste; X11 → xclip; headless → xclip), matching Pillow's original behavior.
  - **Silent-error short-circuit**: If the session-native tool reports a silent error (empty clipboard / no image), returns `None` immediately. This prevents reading stale data from a different clipboard namespace (e.g. XWayland vs Wayland). Real errors and timeouts still fall back to the next backend.
  - **Timeout protection**: `subprocess.run(..., timeout=3)` with `TimeoutExpired` handling prevents the TUI from freezing if a clipboard tool blocks.

### 2. `src/kimi_cli/ui/shell/prompt.py`

- Imports the new `is_media_clipboard_available()` function.
- `Ctrl+V` keybinding is registered when **either** text clipboard (pyperclip) or media clipboard (xclip/wl-paste) is available.
- Text paste fallback inside the `Ctrl+V` handler only runs when pyperclip actually works.
- `PyperclipClipboard` is only created when pyperclip works; otherwise `clipboard = None`, letting `PromptSession` safely fall back to `InMemoryClipboard`.

### 3. `tests/test_clipboard.py`

- **6 tests** for `is_media_clipboard_available()` covering all platform/tool combinations.
- **11 tests** for `_grab_image_linux()` covering:
  - xclip real error → fallback to wl-paste
  - xclip silent error → return None (headless, trusted)
  - both tools silent error → return None after first
  - xclip missing, wl-paste succeeds
  - both tools missing → returns None
  - xclip succeeds immediately → wl-paste not tried
  - **Wayland prefers wl-paste**
  - **Wayland wl-paste real error → fallback to xclip**
  - **Wayland wl-paste silent error → no xclip fallback**
  - **X11 prefers xclip**
  - **subprocess timeout → continues to next backend**

## Edge Cases Considered

| Scenario | Behavior |
|----------|----------|
| macOS / Windows local | **No change** — native APIs work as before |
| Linux desktop (X11) | **No change** — xclip tried first, trusted on silent error |
| Linux desktop (Wayland) | **No change / improved** — wl-paste tried first, trusted on silent error; no stale XWayland data |
| Linux headless + cc-clip (xclip bridge) | **Fixed** — xclip tried first; real errors fall back to wl-paste |
| Linux headless + xclip present, no cc-clip | **Safe** — `Ctrl+V` bound, timeout prevents freeze; built-in keybindings use `InMemoryClipboard` |
| Linux headless + no `xclip`/`wl-paste` | **No change** — both checks return `False` |

## Testing

- [x] Verified `is_media_clipboard_available()` returns `True` on a headless Linux server with `xclip` present
- [x] Verified `Ctrl+V` successfully pastes images when cc-clip tunnel is active
- [x] Verified no crash occurs when `xclip` is unavailable or fails
- [x] Verified prompt_toolkit built-in keybindings (ctrl-k, ctrl-w, ctrl-y) do not crash when pyperclip is broken
- [x] Cross-platform impact reviewed — zero regression on macOS/Windows/Wayland/X11
- [x] All tests pass (32 existing + 17 new)
- [x] `pre-commit run --all-files` passed

## Related

This fix enables seamless clipboard image paste for users who run Kimi CLI on remote Linux machines via SSH, particularly when combined with clipboard bridging tools like cc-clip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Only write entries that are worth mentioning to users.
 
 ## Unreleased
 
-- Shell: Enable clipboard image paste on headless Linux over SSH — when pyperclip is unavailable (e.g. DISPLAY is not set), Ctrl-V now falls back to xclip or wl-paste via PIL so remote clipboard bridges can still inject images; also prevents a UI crash from built-in clipboard shortcuts when pyperclip is broken
+- Shell: Enable clipboard image paste on headless Linux over SSH — when pyperclip is unavailable (e.g. DISPLAY is not set), Ctrl-V now falls back to xclip or wl-paste so remote clipboard bridges can still inject images; also prevents a UI crash from built-in clipboard shortcuts when pyperclip is broken
 
 ## 1.40.0 (2026-04-28)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Only write entries that are worth mentioning to users.
 
 ## Unreleased
 
+- Shell: Enable clipboard image paste on headless Linux over SSH — when pyperclip is unavailable (e.g. DISPLAY is not set), Ctrl-V now falls back to xclip or wl-paste via PIL so remote clipboard bridges can still inject images; also prevents a UI crash from built-in clipboard shortcuts when pyperclip is broken
+
 ## 1.40.0 (2026-04-28)
 
 - Core: Fix `--yolo` mode unintentionally preventing the model from calling `AskUserQuestion` — yolo used to inject a system reminder telling the model it was in "non-interactive mode" and must not ask, and the ask-user tool auto-dismissed in yolo. Both were wrong: yolo only bypasses permission approvals; it does not mean "the user is gone". Yolo no longer injects model guidance, and the user remains reachable through `AskUserQuestion`
--- a/docs/en/release-notes/changelog.md
+++ b/docs/en/release-notes/changelog.md
@@ -4,6 +4,8 @@ This page documents the changes in each Kimi Code CLI release.
 
 ## Unreleased
 
+- Shell: Enable clipboard image paste on headless Linux over SSH — when pyperclip is unavailable (e.g. DISPLAY is not set), Ctrl-V now falls back to xclip or wl-paste via PIL so remote clipboard bridges can still inject images; also prevents a UI crash from built-in clipboard shortcuts when pyperclip is broken
+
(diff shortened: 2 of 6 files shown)


ZichenWen1:
@codex review

ZichenWen1:
@codex review

ZichenWen1:
@claude review

ZichenWen1:
@codex review

PR #2177, opened 2026-05-07
fix(soul): clear partial UI output when LLM step is retried

7Sageer:
## Related Issue

N/A — small targeted bug fix.

## Description

When a streamed LLM call fails after emitting partial content, tenacity retries the step but the aborted attempt's text/think/tool-call output was already shown to the user, so the new attempt appeared concatenated with the failed one.

This PR adds a `StepRetry` wire event emitted from tenacity's `before_sleep` hook so UI consumers can discard the partial state and render a retry banner before the new attempt's output.

- **Live shell view** clears `_current_content_block`, `_tool_call_blocks`, `_last_tool_call_block`, then prints `Retrying after <reason> · attempt N/M · <wait>`. State that is independent of the LLM stream (status, compaction, MCP loading, notifications, approval queues) is intentionally preserved — see the docstring on `_LiveView.discard_retry_attempt` for the rationale.
- **Print backends** (`JsonPrinter`, `FinalOnlyTextPrinter`, `FinalOnlyJsonPrinter`) drop the partial assistant-message buffers; pending notifications captured during the failed attempt still flush so they aren't lost.
- **Web client** (`web/src/hooks/wireTypes.ts`, `useSessionStream.ts`) consumes `StepRetry` for both the parent agent and subagents: the partial thinking/text blocks and any in-flight tool-call messages from the failed attempt are removed, a single live status line `Retrying after … · attempt N/M · <wait>` is rendered (and updated in place across consecutive retries), and tool calls from earlier successful steps in the same turn are preserved (filtered by `tc.result === undefined`).
- **ACP session** receives the event but cannot un-send streamed chunks; addressing the resulting visible duplication is left as a follow-up.

**Wire protocol version is bumped `1.9 → 1.10`** to record the additive `StepRetry` event, matching the convention from prior wire-type changes (#1601, #1552, #1743). The change is backward-compatible — older readers of `wire.jsonl` skip unknown record types via existing exception handling in `WireMessageEnvelope.to_wire_message()` and the replay loop, so no client breaks. Public wire docs (`docs/{en,zh}/customization/wire-mode.md`) are updated with the new event schema.

### Tests
- `test_kimisoul_retry_recovery.py::test_step_retry_event_after_partial_stream` — partial stream → 429 → retry yields only the new attempt in history.
- `test_wire_message.py` — StepRetry serde snapshot.
- `test_empty_think_part_indicator.py::test_step_retry_clears_partial_content_and_updates_live_status` — live view clears blocks and renders banner.
- `test_print_notifications.py` — three new tests covering each print backend's discard behavior.
- Manual smoke: triggered real 429 / connection error and verified the retry banner shows and prior partial output is not duplicated.
- Bonus unintentional smoke: while running `make gen-changelog` for this PR's own version-bump commit, the upstream model returned 503 — the new retry banner rendered correctly (`Retrying after server error · attempt 2/3 · <1s`), confirming `error_type → reason` mapping and attempt/wait formatting on a real failure path.

### Known limitations / follow-ups
- ACP clients still see duplicate content on retry (chunks already streamed cannot be un-sent).
- **vis (`vis/src/features/wire-viewer/*`)** TypeScript client does not yet handle `StepRetry`; it will silently ignore the event, so the wire-viewer will continue to render the failed attempt's partial output until a follow-up PR adds the handler. Web is handled in this PR (see the bullet above).
- `_run_with_connection_recovery` performs first-layer retries (OAuth refresh, single-shot connection recovery) silently, without emitting `StepRetry`.
- No `track("api_retry", ...)` telemetry and no `retries` counter in the session-stats endpoint yet.
- Tools may execute during streaming, so a retry can re-invoke the same tool — pre-existing kosong behavior.

Before:
截屏2026-05-07 22 15 33

After:
截屏2026-05-07 22 17 19

## Checklist

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2177" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


7Sageer:
@codex

7Sageer:
@codex

PR #2198, opened 2026-05-08
fix(acp): defer available commands update to prevent race condition

nelsonBlack:
## Related Issue

No related issue — this is a small bugfix for a race condition in v1.41.0 where slash commands fail to appear in the CLI.

## Description

In version 1.41.0, the ACP server sends the available_commands_update immediately during new_session(), before the client has finished creating its thread view. This causes a race condition where slash commands never appear.

This PR defers the session_update call by wrapping it in an async task with await asyncio.sleep(0), yielding control so new_session() returns first. It also captures self.conn in a local variable before the closure to satisfy pyright type narrowing.

## Changes
- Wrap session_update(available_commands_update) in _send_available_commands_later() coroutine
- Add conn = self.conn before the closure to avoid reportOptionalMemberAccess in pyright
- Use asyncio.sleep(0) to defer until the client is ready
- Update CHANGELOG.md Unreleased section
- Add integration test: test_new_session_sends_available_commands_update

## Checklist

- [x] I have read the CONTRIBUTING document.
- [ ] I have linked the related issue, if any. (No existing issue for this specific race condition)
- [x] I have run make gen-changelog to update the changelog. (Updated CHANGELOG.md directly)
- [ ] I have run make gen-docs to update the user documentation. (No user-facing docs changes needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Only write entries that are worth mentioning to users.
 
 ## Unreleased
 
+- ACP: Fix slash commands not appearing in the CLI due to a race condition during session initialization — the available commands update is now deferred until after `new_session()` returns
+
 ## 1.41.0 (2026-04-30)
 
 - Plugin: Support installing plugins directly from a `.zip` URL — `kimi plugin install` now accepts HTTP(S) URLs ending in `.zip` (e.g. GitHub/GitLab archive links like `.../archive/refs/heads/main.zip`) and downloads + extracts them before resolving `plugin.json`, in addition to the existing git URL, local directory, and local zip-file sources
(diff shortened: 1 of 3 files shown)


nelsonBlack:
@Copilot All three review comments are addressed in the latest commit:

1. **Fire-and-forget task discarded** — Added `self._pending_tasks: set[asyncio.Task[Any]]` to ACPServer. The task is now stored in the set, with a `done_callback` to discard it when complete. Follows the same pattern as `HookEngine.fire_and_forget_trigger` and `background/agent_runner.py`.

2. **Unobserved exceptions** — Added `ACPServer._log_available_commands_failure` as a staticmethod done_callback. It checks `task.cancelled()` and logs any exception via `logger.opt(exception=exc).warning(...)`, consistent with other fire-and-forget loggers in the codebase.

3. **self.conn type narrowing** — Captured `conn = self.conn` before the closure (after the assert) and use `conn.session_update(...)` inside it. This satisfies pyright strict mode.

All checks pass (ruff + pyright) and the integration test passes.

PR #2246, opened 2026-05-12
feat(shell): add --prompt-interactive option

### LINKED ISSUE #2240: Feature Request: Add an option to pass initial prompt while keeping interactive mode
## Current Behavior

Currently, the `--prompt` (or `-p`) option is designed for **single-shot execution**: CLI processes the prompt and then exits immediately without entering the interactive shell.

```sh
kimi -p "Explain this codebase"   # executes and exits
```

This is confirmed by the [documentation](https://moonshotai.github.io/kimi-cli/en/reference/kimi-command.md#input-and-commands):
> "When using `--prompt` (or `--command`), Kimi Code CLI exits after processing the query."

## Desired Behavior

It would be very useful to have an option (e.g. `--prompt-interactive`, `--init-prompt`, or a special value for `--prompt`) that allows users to **pass an initial user message when launching the CLI, but then keep the interactive session open** for follow-up questions.

For example:

```sh
# Hypothetical usage
kimi --prompt-interactive "Analyze the dependency graph of src/main.js"

# Or
kimi -p "Analyze the dependency graph of src/main.js" --interactive
```

After the AI responds to the initial prompt, the user can continue typing follow-up questions in the same session.

## Use Cases

1. **Shell aliases / wrapper scripts**: Users can create shortcuts like `kimi-code-review` that automatically seed a specific prompt but still allow interactive refinement.
2. **IDE integrations**: External editors could launch Kimi with a pre-filled context (e.g. current file, selected lines) and let the developer continue the conversation.
3. **Session initialization**: Quickly start a session with a complex multi-line prompt without having to paste it manually after the shell starts.

## Possible Implementation Ideas

- Add a new flag like `--prompt-interactive TEXT` or `--init-prompt TEXT`.
- Alternatively, make `--prompt` behavior depend on whether `--print` is present: without `--print`, enter interactive mode after handling the initial prompt. (However, this would be a breaking change.)
- Or support reading from a special file descriptor / here-document that gets injected as the first user message while keeping stdin open for the interactive TTY.

## Workarounds Tried

- Custom Agent with `system_prompt_path` works for **fixed** context, but not for dynamic initial prompts.
- Two-step process (`kimi -S session -p "..."` then `kimi -C`) works but is cumbersome and splits the conversation into two commands.
- Piping via `echo "..." | kimi` does not work because the CLI detects non-TTY stdin and skips interactive mode.

Thanks for considering this!

shuizhongyueming:
## Related Issue

Resolve #2240

## Description

This PR adds a new CLI option `--prompt-interactive` (short: `-P`) that allows users to pass an initial prompt when starting the shell UI, and then keeps the interactive session open for follow-up questions.

### Current behavior

The existing `--prompt` (or `-p`) option is designed for single-shot execution: the CLI processes the prompt and exits immediately.

### New behavior

With `--prompt-interactive`, the CLI:
1. Processes the initial prompt
2. Then enters the normal interactive shell loop

### Changes

- **CLI (`src/kimi_cli/cli/__init__.py`)**: Added `--prompt-interactive`/`-P` parameter, mutually exclusive with `--prompt`, only available in shell UI mode.
- **App (`src/kimi_cli/app.py`)**: Passes `initial_command` through to `Shell.run()`.
- **Shell (`src/kimi_cli/ui/shell/__init__.py`)**: Executes `initial_command` before entering the `while True` interactive loop, with proper `resume_prompt` management to pause user input during initial command execution.
- **Tests**: Added unit tests for `initial_command` behavior and E2E tests for CLI validation (mutual exclusion, UI restriction, empty prompt).

### Usage example

```bash
kimi --prompt-interactive "Analyze the dependency graph of src/main.js"
```

After the AI responds to the initial prompt, the user can continue typing follow-up questions in the same session.

## Checklist

### DIFF
--- a/src/kimi_cli/app.py
+++ b/src/kimi_cli/app.py
@@ -684,7 +684,11 @@ class KimiCLI:
                         await external_cancel_task
 
     async def run_shell(
-        self, command: str | None = None, *, prefill_text: str | None = None
+        self,
+        command: str | None = None,
+        *,
+        prefill_text: str | None = None,
+        initial_command: str | None = None,
     ) -> bool:
         """Run the Kimi Code CLI instance with shell UI."""
         from kimi_cli.ui.shell import Shell, WelcomeInfoItem
@@ -767,7 +771,7 @@ class KimiCLI:
         )
         async with self._env():
             shell = Shell(self._soul, welcome_info=welcome_info, prefill_text=prefill_text)
-            return await shell.run(command)
+            return await shell.run(command, initial_command=initial_command)
 
     async def run_print(
         self,
--- a/src/kimi_cli/cli/__init__.py
+++ b/src/kimi_cli/cli/__init__.py
@@ -606,7 +606,7 @@ def kimi(
                 session = await Session.create(work_dir)
                 logger.info("Created new session: {session_id}", session_id=session.id)
 
-            nonlocal _latest_created_session
+            nonlocal _latest_created_session, prompt_interactive
             _latest_created_session = session
 
             # Add CLI-provided additional directories to session state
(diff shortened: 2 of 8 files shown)


shuizhongyueming:
@copilot review

PR #2260, opened 2026-05-13
feat: add kill_ring_system_clipboard config option

sdkks:
## Related Issue

N/A — This change is under 100 lines (per [CONTRIBUTING.md](https://github.com/MoonshotAI/kimi-cli/blob/main/CONTRIBUTING.md)), so no prior issue discussion is required.

## Description

Add a `kill_ring_system_clipboard` configuration option (default: `true`) that allows users to prevent emacs-style kill commands from polluting the system clipboard.

**Problem:** Kimi CLI uses `prompt_toolkit`'s `PyperclipClipboard`, which means all built-in emacs "kill" commands (`Ctrl-W`, `Ctrl-K`, `Ctrl-U`, `Ctrl-Delete`, `Alt-D`, `Alt-Backspace`) write deleted text to the system clipboard via `pyperclip`. This silently destroys existing clipboard content and pollutes clipboard history managers.

**Solution:** When `kill_ring_system_clipboard = false`, eager keybinding overrides intercept the 6 kill commands before prompt_toolkit's built-in handlers can touch the clipboard. The text is deleted without being copied. The in-memory kill ring still works for `Ctrl-Y` within the same session.

Intentional cut/copy/paste operations (`Ctrl-W` with text selected, `Alt-W`, `Ctrl-Y`, `Ctrl-V`) continue to use the system clipboard regardless of this setting.

**Files changed:**
- `src/kimi_cli/config.py`: Added `kill_ring_system_clipboard: bool = Field(default=True, ...)`
- `src/kimi_cli/ui/shell/__init__.py`: Pass config value into `CustomPromptSession`
- `src/kimi_cli/ui/shell/prompt.py`: Add conditional eager keybinding overrides for the 6 kill commands
- `docs/en/configuration/config-files.md`: Document the new option

## Checklist

- [x] I have linked the related issue, if any. (N/A — under 100 lines)
- [x] I have added tests that prove my fix is effective or that my feature works. (Keybinding UI behavior is not currently unit-tested in this codebase; the change is verified by prek format/check pass and manual code review against prompt_toolkit's original handlers.)

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2260" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Only write entries that are worth mentioning to users.
 
 ## Unreleased
 
+- Config: Add `kill_ring_system_clipboard` option — set to `false` to keep emacs-style kill commands (`Ctrl-W`, `Ctrl-K`, `Ctrl-U`, etc.) in an in-memory kill ring instead of overwriting the system clipboard
+
 ## 1.43.0 (2026-05-12)
 
 - Security: Bump pillow to 12.2.0 to address CVE-2026-25990 (out-of-bounds write when loading PSD images); unblocks installs in environments that gate on the older pinned version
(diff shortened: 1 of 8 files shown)


sdkks:
**Author response to review bot comments:**

The core design choice here is intentional: when `kill_ring_system_clipboard = false`, killed text is **discarded** — not saved to any clipboard or ring. The config description was misleading (said "in-memory ring") and has been corrected to "deleted text is discarded." The purpose is to give users who want zero clipboard pollution a way to delete text without it being retained anywhere.

Specific responses:

1. **Selection-cut with Ctrl-W** — Fixed. Added `filter=~has_selection` to the `c-w` override so the built-in emacs cut handler takes over when text is selected. The override now only fires when no selection is active.

2. **"In-memory kill ring" not implemented** — This is by design. There is no kill ring because the flag's purpose is to prevent killed text from being retained *anywhere*. Users who want kill/yank should leave the default (`true`).

3. **Ctrl-Y yanking stale content** — Expected behavior when the flag is off. Ctrl-Y reads from the system clipboard; if kills don't write to it, there's nothing fresh to yank. This is the trade-off of the setting.

PR #2324, opened 2026-05-19
fix(web): handle BrokenPipeError in SessionProcess.send_message

Ricardo-M-L:
## What does this PR do?

`SessionProcess.send_message` in `src/kimi_cli/web/runner/process.py` writes to `process.stdin` and awaits `drain()` without guarding against the subprocess having exited between the `start()` call at the top of the method and the actual write at the bottom:

```python
process.stdin.write((message + "\n").encode("utf-8"))
await process.stdin.drain()
```

In a normal lifecycle `_read_loop` observes the exit and emits a `"stopped"` / `"crashed"` status. But there is a window — subprocess crashes, hits OOM, or is killed by an external signal — where `send_message()` reaches the write before `_read_loop` runs. The result is a raw `BrokenPipeError` (or `ConnectionResetError` on some platforms) propagating up to the FastAPI / websocket handler, surfacing as a 500 to the client with no `error` status emitted to any attached websocket subscribers.

## Fix

Wrap the write + drain pair in a try/except for `BrokenPipeError` and `ConnectionResetError`, log a warning with `process.returncode`, and emit an `"error"` status with `reason="stdin_broken"` so subscribers see the failure synchronously — instead of waiting on `_read_loop` to eventually emit a different terminal status.

No behavior change on the happy path. 14-line addition inside the existing function.

## Repro

In a long-running web session, manually `kill -9` the subprocess (or trigger an OOM in it) and immediately send a message. Without this PR, the next `send_message()` call raises and the websocket clients see only `_read_loop`'s eventual status; the API call itself crashes with a 500.

## Test plan
- [x] `python3.11 -c "import ast; ast.parse(open('.../process.py').read())"` — passes
- [x] Manual code review for happy-path equivalence — the original two-line write/drain is preserved unchanged inside the `try`

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2324" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/kimi_cli/web/runner/process.py
+++ b/src/kimi_cli/web/runner/process.py
@@ -661,8 +661,30 @@ class SessionProcess:
             logger.error(f"{e.__class__.__name__} {e}: Invalid JSONRPC in message: {message}")
             return
 
-        process.stdin.write((message + "\n").encode("utf-8"))
-        await process.stdin.drain()
+        try:
+            process.stdin.write((message + "\n").encode("utf-8"))
+            await process.stdin.drain()
+        except (BrokenPipeError, ConnectionResetError) as e:
+            # Subprocess died between our `start()` check above and the actual write.
+            # `_read_loop` will eventually observe the exit and emit "stopped" /
+            # "crashed", but right now the caller (FastAPI / websocket handler) would
+            # otherwise see a raw exception propagate to the response. Emit an error
+            # status so any attached websocket clients see the failure synchronously.
+            #
+            # If this was a prompt, its id was already registered in
+            # `_in_flight_prompt_ids` above; drop it so the failed turn doesn't
+            # leave the session wedged in `is_busy` forever.
+            if isinstance(in_message, JSONRPCPromptMessage):
+                self._in_flight_prompt_ids.discard(in_message.id)
+            logger.warning(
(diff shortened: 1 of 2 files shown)


Ricardo-M-L:
Thanks for the review. Addressed the in-flight-prompt issue both bots flagged: on the `BrokenPipeError`/`ConnectionResetError` path, a `JSONRPCPromptMessage` id was registered in `_in_flight_prompt_ids` (and `busy` emitted) before the failed write, so the session stayed wedged in `is_busy`. The error path now drops that id before emitting `stdin_broken`, and I added a regression test in `tests/web/test_session_error_recovery.py` asserting `is_busy` is cleared and the status is `error`/`stdin_broken`.

Also rebased onto current `main` (was 15 commits behind); the source diff is just `web/runner/process.py`. CI should be green now — happy to squash if preferred.

Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

Ricardo-M-L:
Friendly ping - this PR has been open since May. Would appreciate a review. Thanks!

Ricardo-M-L:
Ping! Checking in on this PR.

PR #2452, opened 2026-06-14
fix(tools): fail StrReplaceFile when a multi-edit hunk is unmatched

Osamaali313:
## Problem

`StrReplaceFile` applies edits with `str.replace` and then only errors when the **entire** result is unchanged:

```python
for edit in edits:
    content = self._apply_edit(content, edit)

if content == original_content:
    return ToolError(message="No replacements were made. ...")
```

So in a **multi-edit** call where some edits match and others don't, the non-matching edits are silently skipped and the tool still reports success:

```python
# content = "Hello world! Goodbye world!"
edit = [Edit(old="Hello", new="Hi"), Edit(old="does-not-exist", new="x")]
# -> result.is_error == False
# -> message: "Applied 2 edit(s) with 1 total replacement(s)."
# -> file becomes "Hi world! Goodbye world!"  (edit #2 silently dropped)
```

The model is told all hunks landed when one never did — a silent loss of requested changes. This is easy to hit when the file has drifted from what the model last read (the same situation behind the frequent `old_string not found` confusion), because a partially-stale batch *looks* like it succeeded.

A second, smaller bug: the reported replacement **count** was computed against the *original* content for every edit, so it's wrong for sequential dependent edits (e.g. `foo`→`bar` then `bar`→`baz` reported `1` instead of `2`, since `"bar"` isn't in the original `"foo"`).

## Fix

Apply edits sequentially and validate each edit's `old` string is present **at its turn** (a later edit may legitimately target text an earlier edit introduced). If any edit doesn't match, fail the whole call **without writing**, so a multi-edit is atomic and the failure is surfaced with the offending edit index. The replacement count is now tallied as edits actually apply, fixing the sequential-edit count too.

Behavior preserved:
- Single edit that matches / doesn't match → unchanged (still `"No replacements were made"`).
- Independent multi-edits that all match → unchanged.
- The no-op case (`old == new`) is still guarded.

## Tests

Added `test_replace_multiple_edits_one_not_found_is_atomic` (asserts the batch fails and the file is left untouched) and `test_replace_sequential_dependent_edits_count` (dependent edits apply and the count is correct). Full `tests/tools/test_str_replace_file.py` passes (15), plus the plan-mode/additional-dirs suites (19). `ruff check` / `ruff format` clean.

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2452" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/kimi_cli/tools/file/replace.py
+++ b/src/kimi_cli/tools/file/replace.py
@@ -142,6 +142,20 @@ class StrReplaceFile(CallableTool2[Params]):
             # requested changes without the model noticing.
             total_replacements = 0
             for index, edit in enumerate(edits):
+                # Reject an empty old string: `str.count('')` returns
+                # `len(content) + 1` (never 0), so the occurrence check below
+                # cannot catch it, and `str.replace('', new)` would splice
+                # `new` between every character.
+                if edit.old == "":
+                    detail = (
+                        "old string must not be empty."
+                        if len(edits) == 1
+                        else f"Edit #{index + 1}'s old string must not be empty."
+                    )
+                    return ToolError(
+                        message=f"No replacements were made. {detail}",
+                        brief="Empty old string",
+                    )
                 occurrences = content.count(edit.old)
                 if occurrences == 0:
                     detail = (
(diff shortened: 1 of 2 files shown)


Osamaali313:
Thanks @devin-ai-integration — both are valid. Addressed in 884c671:

1. **No-op message:** when every edit matches but the result equals the original (`old_string == new_string`, or edits that cancel out), the final guard no longer claims "old string was not found". It now reports that the edit(s) matched but left the file unchanged, so the model isn't sent re-reading the file for a string that is actually present.

2. **Empty `old_string`:** added an explicit per-edit guard. As noted, `str.count('')` is `len(content) + 1` (never 0) so the occurrence check can't catch it, and `str.replace('', new)` would splice `new` between every character. Empty `old` is now rejected with a clear message.

Added tests for both (`test_replace_noop_reports_unchanged_not_missing`, `test_replace_rejects_empty_old_string`); full tool suite passes (17), ruff check/format clean.

PR #2471, opened 2026-06-22
feat(tools): add Monitor tool for per-line stdout streaming

Nitjsefnie:
## Related Issue

N/A — feature proposal (no existing issue).

## Description

Adds a **`Monitor`** tool: a streaming counterpart to the existing background-task tools. Where `TaskList` / `TaskOutput` / `TaskStop` are poll-based and the background subsystem notifies **once** on terminal state, `Monitor` runs a self-filtering command and emits a notification for **each matching stdout line** as it appears — for watching logs, polling job state, or tailing a server while the agent keeps working. The shape mirrors Claude Code's `Monitor` tool.

**Approach** — a new `monitor` background-task *kind*, reusing the existing worker, output file, lifecycle tools, and notification pipeline:

- The command self-filters (`grep --line-buffered`, `awk … fflush()`), so the task's **output file is the event stream** — every new line is a line to notify. The **worker is unchanged** (the bash path already covers it).
- `BackgroundTaskManager.reconcile()` (already driven by the print/shell loops) gains `publish_monitor_notifications()`: for each active monitor task it reads new output bytes since a per-task `notify_offset`, splits complete lines (carrying a partial trailing line forward), and publishes one batched `monitor_line` `NotificationEvent` through the existing store + `NotificationWatcher` → agent loop.
- **Volume cap:** a monitor exceeding `max_lines_per_window` (default 200 lines / 5 s) is auto-killed with one `warning` notification.
- Root-agent only and blocked in plan mode, matching the other background tools; creation goes through the approval gate like `Shell`.

**Tool surface**

| param | type / default | meaning |
|---|---|---|
| `command` | str (required) | shell command; each stdout line is an event (self-filter it) |
| `description` | str (required) | shown in every notification |
| `timeout_ms` | int = 300000 (1000–3600000) | kill after deadline; ignored when persistent |
| `persistent` | bool = false | session-length watch (no timeout); stop with `TaskStop` |

**Files** — `background/models.py` (`monitor` kind + `MonitorPayload`), `background/ids.py` (id prefix), `background/manager.py` (`create_monitor_task` + `publish_monitor_notifications`), `tools/monitor/` (the tool + description), `agents/default/agent.yaml` (registration). Worker untouched.

**Notes** — `batch_ms` is carried in the payload but batching is currently effectively per-`reconcile()` pass; a finer time-based coalescer can follow if desired. The 8 unrelated failures in `tests/ui/test_usage.py` / `tests/utils/test_list_directory.py` are pre-existing on `main` (verified by reproducing them on the base commit with no Monitor code).

Design doc (architecture, data flow, risks, alternatives): https://docs.nitjsefni.eu/#/d/kimi-cli/monitor-tool-design

## Checklist

- [ ] I have linked the related issue, if any. 

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Nitjsefnie:
Good catch by Devin Review — fixed in bcc1db0d. `publish_monitor_notifications` now flushes the trailing partial line once a monitor task is terminal (both the "full line + trailing partial" and "single line, no newline ever" cases), instead of waiting for a newline that will never arrive. Added two regression tests (`test_terminal_task_flushes_trailing_partial_line`, `test_terminal_task_flushes_single_unterminated_line`). Running monitors still correctly hold a partial line until its newline arrives.

Nitjsefnie:
Rebased onto `main` — this merges cleanly again (`mergeable` is back to MERGEABLE).

The only conflict was the changelog: `main` cut 1.49.0 and populated the section our `Unreleased` entries sat in. Both sides were wanted, so the Monitor entries stay under `## Unreleased` and the 1.49.0 section is untouched, across `CHANGELOG.md` and both `docs/en` / `docs/zh` changelogs. I diffed the PR's own `base...head` patch before and after the rebase to confirm nothing else moved — identical file set, identical content outside those three files.

While re-running the gates I also found that `ruff format --check` was already failing on this branch *before* the rebase, on `tests/background/test_monitor_notifications.py` — a wrapped signature that fits in the configured 100-char limit. It came from our own earlier commit, and ruff is pinned at the same 0.14.14 on both sides, so this wasn't tooling drift; CI's format gate would have failed regardless. Fixed with ruff's own output.

Gates now: `ruff check` clean, `ruff format --check` → 542 files already formatted, `pyright` → 0 errors, and all 84 Monitor/background/agent tests pass. The remaining suite failures are pre-existing — I ran the same subset against clean `main` in a separate worktree and got 17 failures in the same files (flaky PTY e2e, plus a `chmod 0` test that can't fail meaningfully as uid 0).

Note the rebase rewrote SHAs, so my earlier comment's reference to `bcc1db0d` now points at the pre-rebase commit.

Nitjsefnie:
Closing this — I'm no longer able to keep it current against `main`, and it's better withdrawn than left as a stale rebase target. The Devin finding from the first round (trailing partial line lost when a monitor terminates) was real and is fixed in the branch history if anyone wants to pick the work up.

PR #2493, opened 2026-07-11
Fix: record started_at for background agent tasks so duration is reported

nankingjing:
### Problem

Background **agent** tasks never have `runtime.started_at` set, so their run
duration is silently lost. Background **bash** tasks set it correctly.

The only place `started_at` is ever assigned is the bash worker startup
(`background/worker.py`):

```python
runtime.status = "starting"
runtime.worker_pid = os.getpid()
runtime.started_at = time.time()
```

Agent tasks reach the running state through `BackgroundTaskManager._mark_task_running`
(called from `background/agent_runner.py`), which set every startup field
**except** `started_at`:

```python
runtime.status = "running"
runtime.updated_at = time.time()
runtime.heartbeat_at = runtime.updated_at
runtime.failure_reason = None
```

Since `TaskRuntime.started_at` defaults to `None` (`background/models.py`), agent
tasks keep `started_at is None` for their whole lifetime.

### Impact

Downstream consumers guard on `started_at`, so for agent tasks they silently
degrade:

- `publish_terminal_notifications` computes `duration_s = None`, so the completion
  notification omits the `Duration: …` line that bash tasks show.
- `_mark_task_completed` / `_mark_task_failed` / `_mark_task_timed_out` /
  `_mark_task_killed` all gate telemetry on
  `if runtime.started_at and runtime.finished_at:`, which is never true for agent
  tasks — so the `background_task_completed` event (with `duration_s`) is never
  emitted for them.

### Fix

Record the first transition into `running` as the start time in
`_mark_task_running`, mirroring what the bash worker does. The assignment is
guarded by `started_at is None` because `_mark_task_running` is also invoked
again after each approval is resolved; overwriting would otherwise shrink the
measured duration to the last approval gap.

```python
runtime.status = "running"
runtime.updated_at = time.time()
if runtime.started_at is None:
    runtime.started_at = runtime.updated_at
runtime.heartbeat_at = runtime.updated_at
```

This has no effect on bash tasks (they don't go through `_mark_task_running`) and
is a no-op for tasks that already reached a terminal state (the early return
guarding terminal status is untouched).

### Test

Adds `test_mark_task_running_sets_started_at_once`, asserting that the first
`_mark_task_running` sets `started_at`, a subsequent re-mark (e.g. after an
approval) preserves it, and completion then yields `finished_at >= started_at`.

### Verification

- `python -m py_compile` on both changed files: passes.
- New/existing tests verified by reading; not executed in my environment
  (runtime deps such as `kosong`/`kaos` were not installable here).

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2493" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/kimi_cli/background/manager.py
+++ b/src/kimi_cli/background/manager.py
@@ -616,6 +616,13 @@ class BackgroundTaskManager:
             return
         runtime.status = "running"
         runtime.updated_at = time.time()
+        # Record the first transition into "running" as the task's start time so
+        # that completion duration can be computed. Agent tasks reach the
+        # "running" state here (bash tasks set started_at in the worker); guard
+        # against overwriting it because this method is also called again after
+        # each approval is resolved, which would otherwise reset the duration.
+        if runtime.started_at is None:
+            runtime.started_at = runtime.updated_at
         runtime.heartbeat_at = runtime.updated_at
         runtime.failure_reason = None
         self._store.write_runtime(task_id, runtime)
--- a/tests/background/test_manager.py
+++ b/tests/background/test_manager.py
@@ -826,6 +826,47 @@ def test_mark_task_running_and_completed_clear_approval_reason(runtime):
     assert completed.runtime.failure_reason is None
 
 
+def test_mark_task_running_sets_started_at_once(runtime):
+    manager = runtime.background_tasks
+    store = manager.store
+    spec = TaskSpec(
+        id="a4444444",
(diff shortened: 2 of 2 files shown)


nankingjing:
Reviewed — setting started_at once and guarding against overwrite correctly computes agent task duration. The test covers the approval-re-mark scenario well. Ready for review.

PR #2507, opened 2026-07-18
fix(acp): signal QuestionNotSupported instead of resolving empty answers

### LINKED ISSUE #2495: ACP: AskUserQuestion/QuestionRequest resolves empty — structured questions unusable over ACP
## Summary

In ACP server mode (`kimi acp`), `AskUserQuestion` cannot work: the ACP session resolves every `QuestionRequest` with an empty answer dict, so the model always receives `{"answers": {}, "note": "User dismissed the question without answering."}` — even when the user is present and willing to answer. ACP clients (Paseo, Zed, JetBrains AI Chat, …) therefore can never render the question's options or return a selection.

Observed downstream effect (kimi-code 0.23.6 driven by Paseo over ACP): the tool call surfaces as a generic permission card containing the raw `{"questions": [...]}` JSON with only Reject/Accept buttons — the user has no way to pick an option, and if the card is ignored the agent turn hangs indefinitely.

## Current behavior

`src/kimi_cli/acp/session.py` (~line 211):

```python
case QuestionRequest():
    logger.warning("QuestionRequest is unsupported in ACP session; resolving empty answer.")
    msg.resolve({})
```

The empty resolve is indistinguishable from a genuine user dismissal, so the model gets no signal that the *client* was the problem and may retry the tool.

## Why this matters

- `AskUserQuestion` is the documented way for the agent to collect decisions mid-task; over ACP it is silently broken.
- The Wire protocol already solves this (`QuestionRequest` + `supports_question` capability negotiation in `initialize`); ACP has no equivalent path at all.
- The `ask_user` tool already contains a graceful fallback for unsupported clients (`except QuestionNotSupported` → ToolError: *"Do NOT call this tool again. Ask the user directly in your text response instead."*), but the ACP layer never raises it.

## Proposals

**Short-term (one line, helps every ACP client immediately):**

```python
case QuestionRequest():
    msg.set_exception(QuestionNotSupported())
```

This routes into the existing ToolError fallback, so the model asks its question in plain text — far better than a phantom dismissal or a hang.

**Full support (either or both):**

1. Map `QuestionRequest` to `session/request_permission`, encoding each question option as a permission option (option id = label, kind `allow_once`), and translate the selected option id back into `answers`. Covers the common single-question case; multi-question could be sequential requests. Multi-select has no clean mapping.
2. Add an ACP extension method (e.g. `_kimi/questionRequest`) carrying the full `QuestionItem[]` payload with a structured response, so capable clients can render multi-question / multi-select panels; clients advertise support via an extension capability.

## References

- ACP session handling: https://github.com/MoonshotAI/kimi-cli/blob/main/src/kimi_cli/acp/session.py
- `ask_user` tool (empty-answer and `QuestionNotSupported` paths): https://github.com/MoonshotAI/kimi-cli/blob/main/src/kimi_cli/tools/ask_user/__init__.py
- `QuestionRequest` / `QuestionNotSupported` definitions (`resolve` / `set_exception`): https://github.com/MoonshotAI/kimi-cli/blob/main/src/kimi_cli/wire/types.py
- Downstream client-side companion issue (Paseo rendering of ACP questions): https://github.com/getpaseo/paseo/issues/2048

ayaangazali:
## Related Issue

Resolve #2495

## Description

In ACP server mode every `QuestionRequest` gets resolved with an empty dict (`src/kimi_cli/acp/session.py` line 211), which is indistinguishable from the user actually dismissing the question. The model receives "User dismissed the question without answering" even though the ACP client never rendered anything, so it can retry the tool or just give up on getting an answer.

This applies the short term fix from the issue: signal `QuestionNotSupported` instead. It's the same thing the Wire server already does for clients without the `supports_question` capability (`src/kimi_cli/wire/server.py` line 1055), and it routes into the existing fallback in the ask_user tool which tells the model to ask the question in plain text instead. So over ACP (Zed, JetBrains, etc) the question actually reaches the user now, just as regular text.

One side effect worth calling out: `EnterPlanMode`/`ExitPlanMode` also send `QuestionRequest`, so over ACP they now return their "client does not support plan mode" ToolError instead of a phantom dismissal. That matches how Wire clients without question support already behave, so I think it's the right call, but flagging it in case you see it differently.

Follow up ideas from the issue (out of scope here): hide the AskUserQuestion tool in ACP mode like `_sync_ask_user_tool_visibility` does for Wire, or map questions onto `session/request_permission`. Happy to take a stab at either if there's interest.

Test: added `tests/acp/test_session_questions.py` which runs a prompt turn through `ACPSession` with a fake CLI yielding a `QuestionRequest` and asserts the request resolves with `QuestionNotSupported`.

## Checklist

- [x] I have run `make gen-changelog` to update the changelog. (i updated CHANGELOG.md by hand in the same style, the gen-changelog skill needs a Kimi API setup i don't have)
- [x] I have run `make gen-docs` to update the user documentation. (checked docs/, nothing describes ACP question behavior so there was nothing to regenerate)

---

small disclosure: i'm a freshman in college and i talked the logic through with claude code before writing this. the diff is tiny but if i misread how any of this works i would honestly love to know, still learning :)

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2507" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Only write entries that are worth mentioning to users.
 ## Unreleased
 
 - Kosong: Stop sending an empty `anthropic-beta` header when no beta features are declared — adaptive thinking removes the interleaved-thinking beta, which previously left an empty header value that some backends reject
+- ACP: `AskUserQuestion` now signals `QuestionNotSupported` instead of resolving an empty answer, so the model falls back to asking in plain text rather than seeing a phantom user dismissal
 
 ## 1.49.0 (2026-07-16)
 
--- a/src/kimi_cli/acp/session.py
+++ b/src/kimi_cli/acp/session.py
@@ -30,6 +30,7 @@ from kimi_cli.wire.types import (
     MCPLoadingEnd,
     Notification,
     PlanDisplay,
+    QuestionNotSupported,
     QuestionRequest,
     StatusUpdate,
     SteerInput,
@@ -209,10 +210,13 @@ class ACPSession:
                     case ToolCallRequest():
                         logger.warning("Unexpected ToolCallRequest in ACP session: {msg}", msg=msg)
                     case QuestionRequest():
+                        # Signal the tool that the client cannot render questions,
+                        # so it can tell the LLM to ask in plain text instead.
                         logger.warning(
-                            "QuestionRequest is unsupported in ACP session; resolving empty answer."
+                            "QuestionRequest is unsupported in ACP session; "
+                            "signaling QuestionNotSupported."
                         )
-                        msg.resolve({})
+                        msg.set_exception(QuestionNotSupported())
                     case _:
                         pass
         except LLMNotSet as e:
--- a/tests/acp/test_session_questions.py
+++ b/tests/acp/test_session_questions.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+from typing import Any
+
+import acp
+import pytest
+
+from kimi_cli.acp.session import ACPSession
+from kimi_cli.wire.types import (
+    QuestionItem,
+    QuestionNotSupported,
+    QuestionOption,
+    QuestionRequest,
+    TextPart,
+    TurnBegin,
+    TurnEnd,
+)
+
+
+class _FakeConn:
+    def __init__(self) -> None:
+        self.updates: list[tuple[str, Any]] = []
+
+    async def session_update(self, session_id: str, update: object) -> None:
+        self.updates.append((session_id, update))
+
+
+def _make_question_request() -> QuestionRequest:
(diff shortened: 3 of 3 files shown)


ayaangazali:
Hi, sorry to bump this. Noticed the CI workflows here are sitting at `action_required`, which I think means they need a maintainer to approve runs for a first time contributor, so the checks have never actually executed on this branch. Flagging in case that is easy to click.

For what it is worth I ran the gates locally and they pass (`make check-kimi-cli` and `make test-kimi-cli`). Still merges clean too. Happy to rebase or drop this entirely if it is not something you want, no hard feelings either way.

Issue #2049, opened 2026-04-24
session history lost on resume despite appearing on screen


CyberAttackr:
### What version of Kimi Code CLI is running?

kimi, version 1.36.0

### Which open platform/subscription were you using?

kimi code

### Which model were you using?

kimi-for-coding

### What platform is your computer?

windows 10

### What issue are you seeing?

this is after i run /sessions and resume the session. the prior messages appear on the screen but the agent doesnt have it. I prompted it "do you see the previous messages? ```[previous message]```" and it replied:

"  The user is asking if I see previous messages. I should be honest that I don't see previous conversation history in this session, but I can look at the codebase to understand what w
  as done."

### What steps can reproduce the bug?

create session, fork it a few times, some undos, and then resume it 

### What is the expected behavior?

agent has messages

### Additional information

this is pretty bad

PR #2520, opened 2026-07-19
fix(session): align fork/undo context truncation to wire turns

### LINKED ISSUE #1974: The undo command lands effectively 4 turns earlier than expected

1.37

/login

kimi-for-coding

WSL2 x86_64

When undo command is used and let's say the turn 12 is chosen, what's happening is that effectively the harness is taking the conversation back to turn 8. **This issue might be due to failed requests as well.** I gave the example of lag by 4 but at another place it's lag by 3, with it turning to 4 right after a user message which errored (LLM Provider Error) out.

1. Have a 10-12 turns conversation
2. Use undo to go to turn number 9
3. Conversation actually goes back to turn 5

undoing to turn 9 should take to turn 9.

### Additional information

_No response_

### LINKED ISSUE #2517: /undo and /fork truncate context.jsonl at the wrong turn in compacted or steered sessions

1.49.0 (reproduced on current `main`, commit 4a550ef)

Local OpenAI-compatible mock provider (bug is provider-independent; it is in local session-file handling)

n/a (any)

macOS (Darwin arm64); the code path is platform-independent

`truncate_context_at_turn()` in `src/kimi_cli/session_fork.py` maps a **wire** turn index onto **context.jsonl** by counting non-checkpoint `role == "user"` records. But the context file is not 1:1 with wire turns:

1. **Compaction** rewrites all earlier turns into a single `role="user"` summary record (`compaction.py`: `Message(role="user", content=[system(COMPACTION_OUTPUT_PREFIX), ...])`).
2. **Steers** append extra `role="user"` records mid-turn (`kimisoul._consume_pending_steers`).
3. **Notifications** are injected as `role="user"` records (`notifications/llm.py`).
4. **Slash-command turns** (e.g. `/compact`) produce a wire `TurnBegin` but no context user record.

After any of these, the user-record count no longer matches the wire turn index, so `/undo` and `/fork` cut `context.jsonl` at the wrong record:

- **Compacted session + `/undo`**: the "undone" turn silently **stays in the model's context** while the UI no longer shows it — the agent acts on a turn the user believes was removed.
- **Steered session + `/undo`/`/fork`**: truncation stops **too early** — the forked context is missing turns that the resumed UI (wire replay) still displays. This matches the symptom reported in #2049 ("prior messages appear on the screen but the agent doesn't have it" after forks/undos).

With any provider (below reproduced with a local OpenAI-compatible mock so it is fully deterministic):

1. `kimi --print -p "turn one"` ; `kimi --print -C -p "turn two"` ; `kimi --print -C -p "/compact"` ; `kimi --print -C -p "turn three"`
2. The session now has wire turns `[turn one, turn two, /compact, turn three]` and context user records `[<compaction summary>, turn two, turn three]`.
3. Fork/undo at turn index 2 (i.e. rewind to before "turn three") — via `/undo` in the TUI, `POST /sessions/{id}/fork`, or directly `fork_session(session_dir, work_dir, turn_index=2)`.
4. Inspect the forked session:
   - `wire.jsonl`: turns 0–2 only — "turn three" is gone from the UI. ✅
   - `context.jsonl`: **still contains the "turn three" user message and its assistant reply.** ❌

Observed on main:

```
== forked wire turns ==        == forked context ==
turn 0: 'turn one'             ... user  '<system>Previous context has been compac...'
turn 1: 'turn two'             ... user  'turn two'
turn 2: '/compact'             ... assistant 'hello from mock'
                               ... user  'turn three'   <-- should have been dropped
                               ... assistant 'hello from mock'
```

The forked/undone session's context should contain exactly the turns that remain visible in the wire log — nothing more (compaction case) and nothing less (steer case).

### Additional information

Root cause: `truncate_wire_at_turn` counts `TurnBegin` records while `truncate_context_at_turn` counts user-role records; synthetic user-role records (compaction summary, steers, notifications) and context-less turns (slash commands) break the correspondence in both directions.

**Relation to existing reports:** #1974 and open PR #2386 cover one direction of this desync — wire-only slash-command turns shifting the undo cut point. This report is about the other, unreported direction: **synthetic user-role records in the context** (compaction summaries, steers, notifications). #2386's wire→context index mapping does not exclude these records, so compacted or steered sessions still truncate at the wrong place (and the compaction case leaks "undone" turns into the model's context, which is worse than landing on the wrong turn).

Proposed fix (PR attached): align context user records to wire turns by matching the extracted user text (turns consumed in order; records that match no remaining wire turn — summaries, steers, notifications — do not advance the turn counter). Positional counting is kept as a fallback when wire texts are unavailable. A more invasive alternative would be to write an explicit `_turn` marker record into `context.jsonl` at each turn begin; happy to go that way instead if you prefer a protocol-level fix.

Nas01010101:
## Related Issue

Resolve #2517

Also resolves #1974 (wire-only slash turns shifting the undo cut — covered by a dedicated regression test) and likely the root cause of #2049 (history mismatch after forks/undos).

**Relation to open PR #2386:** that PR maps wire turns to context turns for the slash-command case (credit to its author for identifying that direction). It does not handle synthetic user-role records in the context — compaction summaries, steers, notifications — so compacted/steered sessions still cut at the wrong record, and the compaction case leaks "undone" turns into the model's context. This PR covers both directions with tests for each; happy to instead rebase the additional coverage on top of #2386 if the maintainers prefer to land that one first.

## Description

`truncate_context_at_turn()` counts non-checkpoint `role="user"` records to find the cut point for `/undo` and `/fork`, but the context file is not 1:1 with wire turns: compaction rewrites earlier turns into one user-role summary record, steers and notifications inject extra user-role records, and slash-command turns produce no context record at all. After any of these, the context is truncated at the wrong record:

- rewinding past a post-compaction turn **keeps that turn in the model's context** while the UI no longer shows it;
- forking a steered session cuts **too early**, so the agent is missing turns the resumed UI still displays (the #2049 symptom).

**Fix:** when the wire turn texts are available (from `enumerate_turns`, which `fork_session` already computes for wire truncation), align context user records to wire turns by matching extracted user text, consuming turns in order. Records that match no remaining wire turn (summaries, steers, notifications) no longer advance the turn counter, and turns with no context record (slash commands, compacted-away turns) no longer consume a user record. Positional counting is kept as an explicit fallback when wire texts are unavailable, preserving the existing best-effort semantics.

## Real-behavior proof (deterministic, local mock provider)

Build a real session: `turn one` → `turn two` → `/compact` → `turn three`; then `fork_session(dir, work_dir, turn_index=2)` (the exact `/undo` code path):

- on `main`: forked wire shows turns 0–2, but forked `context.jsonl` **still contains `user: "turn three"` and its assistant reply** ❌
- on this branch: forked context ends after the `turn two` block ✅

## Verification

- New tests in `tests/core/test_session_fork.py::TestTruncateContextAlignment` (4 fail on `main`):
  compaction summary not counted as a turn; steer not counted; duplicate turn texts align in order; end-to-end `fork_session` after compaction drops the undone turn; positional fallback unchanged.
- `uv run pytest tests/core/test_session_fork.py` → 34 passed; full `tests/` suite green.
- `ruff` / `pyright` clean.

## Checklist

- [x] I have read the CONTRIBUTING document.
- [x] I have linked the related issue (file the draft issue first).
- [x] I have added tests that prove my fix is effective.
- [x] Changelog updated (manual `## Unreleased` entry; `make gen-changelog` needs kimi API auth).
- [x] `make gen-docs` — N/A: no user-facing docs change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Only write entries that are worth mentioning to users.
 
 ## Unreleased
 
+- Fixed `/undo` and `/fork` truncating the conversation context at the wrong turn in sessions that had been compacted or steered, which could silently keep turns the user rewound past (or drop turns they kept).
+
 ## 1.49.0 (2026-07-16)
 
 **Highlights**: The completion-token budget for Kimi providers now adapts to the model's remaining context window, reducing context-length overflow errors on long turns
--- a/src/kimi_cli/session_fork.py
+++ b/src/kimi_cli/session_fork.py
@@ -195,6 +195,15 @@ def truncate_context_at_turn(
     Unlike wire truncation, this is best-effort: if context has fewer user turns
     than ``turn_index`` (e.g. slash-command turns that did not mutate context),
     return all available context lines instead of failing.
+
+    Known limitations of text alignment (both need an explicit turn-marker
+    record in context.jsonl to resolve, and both predate this alignment):
+
+    - A context-mutating slash turn (e.g. ``/init``) injects a user record
+      whose text differs from the wire turn text, so it cannot be attributed
+      to its (possibly cut) turn and is conservatively kept.
+    - A steer whose text is exactly equal to a later turn's user text is
+      indistinguishable from that turn's start and aligns to it.
     """
     if not context_path.exists():
         return []
--- a/src/kimi_cli/session_fork.py
+++ b/src/kimi_cli/session_fork.py
@@ -10,6 +10,7 @@ import json
 import mimetypes
 import re
 import shutil
+from collections.abc import Sequence
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, cast
@@ -169,11 +170,27 @@ def _is_checkpoint_user_message(record: dict[str, Any]) -> bool:
     return False
 
 
-def truncate_context_at_turn(context_path: Path, turn_index: int) -> list[str]:
+def truncate_context_at_turn(
+    context_path: Path,
+    turn_index: int,
+    turn_texts: Sequence[str] | None = None,
+) -> list[str]:
     """Read context.jsonl and return all lines up to and including the given turn.
 
     Turn detection is based on real user messages, excluding synthetic checkpoint
     user entries like ``<system>CHECKPOINT N</system>``.
+
+    ``turn_index`` counts *wire* turns, but the context file is not 1:1 with the
+    wire: compaction rewrites earlier turns into a single user-role summary,
+    steers and notifications append extra user-role messages, and slash-command
+    turns never touch the context at all. When ``turn_texts`` (the user text of
+    every wire turn, index-aligned, from :func:`enumerate_turns`) is provided,
+    context user records are aligned to wire turns by matching text, and records
+    that match no wire turn do not advance the turn counter. Plain positional
+    counting — kept as the fallback when ``turn_texts`` is None — treats every
+    user record as a turn boundary and truncates at the wrong record after any
+    compaction or steer, silently keeping turns the user rewound past (or
+    dropping turns they kept).
 
     Unlike wire truncation, this is best-effort: if context has fewer user turns
     than ``turn_index`` (e.g. slash-command turns that did not mutate context),
@@ -184,6 +201,7 @@ def truncate_context_at_turn(context_path: Path, turn_index: int) -> list[str]:
 
     lines: list[str] = []
     current_turn = -1  # Will become 0 on first real user message
+    next_wire_turn = 0  # Alignment cursor into turn_texts
 
     with open(context_path, encoding="utf-8") as f:
         for line in f:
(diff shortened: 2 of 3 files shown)


Nas01010101:
For reviewer transparency — two edge cases the text-alignment approach cannot fully resolve (now documented in the `truncate_context_at_turn` docstring, pushed as a docs commit):

1. **Context-mutating slash turns** (e.g. `/init` appends a user-role instruction record): the injected record's text differs from the wire turn text (`"/init"`), so it cannot be attributed to its turn and is conservatively **kept** even when that turn is rewound past.
2. **Exact-text collision**: a steer whose text is exactly equal to a later turn's user text is indistinguishable from that turn's start and aligns to it, cutting slightly early.

Both predate this PR — positional counting handled them strictly worse (every compaction/steer/slash turn corrupted the cut, not just these collisions) — and both are inherently ambiguous for any content-based heuristic: I checked LCS/sequence alignment (`difflib`) and it resolves the collision case identically, and substring-matching slash injections trades a rare false-negative for likelier false-positives on user text that merely mentions a slash command.

The clean resolution for both is the explicit `_turn` marker record in `context.jsonl` proposed as the protocol-level option in #2517 — happy to implement that as a follow-up if you'd prefer it over (or on top of) this alignment.

PR #2524, opened 2026-07-20
fix(tools): count StrReplaceFile replacements against the running content

### LINKED ISSUE #2526: StrReplaceFile reports too few total replacements for chained edits

`StrReplaceFile` applies its edits sequentially, but it counts the total number of replacements against the *original* file content instead of the running (progressively edited) content. When one edit's `old` string is text produced by an earlier edit, that edit is not present in the original content, so it is counted as zero even though it is applied correctly.

The file is written correctly; only the "total replacement(s)" number in the tool result message is wrong.

### Steps to reproduce

Apply two chained edits to a file containing `hello world`:

    edits = [
        Edit(old="hello",   new="goodbye"),
        Edit(old="goodbye", new="farewell"),
    ]

The file correctly becomes `farewell world`, but the tool reports:

    File successfully edited. Applied 2 edit(s) with 1 total replacement(s).

The second edit (`goodbye` -> `farewell`) is counted as 0 because `goodbye` does not appear in the original content, only in the content produced by the first edit.

### Expected behavior

The reported count should reflect the replacements actually performed, i.e. `2 total replacement(s)` for the example above. Each edit should be counted against the content it is applied to (the running content), not the original.

### Additional information

Found by reading the tool source. I have a fix ready with a regression test.

Sreekant13:
## Related Issue

Resolves #2526

## Description

`StrReplaceFile` applies its edits sequentially, but it computed the reported replacement count against the *original* file content. A chained edit whose `old` string is produced by an earlier edit is not present in the original, so it was counted as zero.

For example, on a file containing `hello world`:

    edits = [
        Edit(old="hello",   new="goodbye"),
        Edit(old="goodbye", new="farewell"),
    ]

both edits replace text and the file correctly becomes `farewell world`, but the tool reported:

    File successfully edited. Applied 2 edit(s) with 1 total replacement(s).

because `"goodbye"` was not found in the original content. This counts each edit against the content it is actually applied to (the running content), so the reported number matches what happened (`2 total replacement(s)` above).

The file output itself is unchanged; only the reported count is corrected. A regression test is added that fails on the current code.

## Checklist

- [x] I have linked the related issue, if any. (#2526)
- [x] I have run `make gen-changelog` to update the changelog. (Entry added to the Unreleased section by hand in the existing style; `make gen-changelog` drives Kimi CLI itself, which I could not run locally.)
- [ ] I have run `make gen-docs` to update the user documentation. (N/A: no user-facing docs change.)

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2524" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Only write entries that are worth mentioning to users.
 ## Unreleased
 
 - Kosong: Stop sending an empty `anthropic-beta` header when no beta features are declared — adaptive thinking removes the interleaved-thinking beta, which previously left an empty header value that some backends reject
+- Tool: Fix StrReplaceFile reporting the wrong replacement count when edits are chained. The total was counted against the original file content, so a later edit whose target text was produced by an earlier edit was counted as zero; the count now tracks the running content
 
 ## 1.49.0 (2026-07-16)
 
--- a/src/kimi_cli/tools/file/replace.py
+++ b/src/kimi_cli/tools/file/replace.py
@@ -134,9 +134,18 @@ class StrReplaceFile(CallableTool2[Params]):
             original_content = content
             edits = [params.edit] if isinstance(params.edit, Edit) else params.edit
 
-            # Apply all edits
+            # Apply all edits, counting replacements against the running content.
+            # Edits apply sequentially, so counting against the original content
+            # miscounts chained edits (e.g. one edit whose `old` is produced by an
+            # earlier edit is not found in the original and would be counted 0).
+            total_replacements = 0
             for edit in edits:
+                before = content
                 content = self._apply_edit(content, edit)
+                if edit.replace_all:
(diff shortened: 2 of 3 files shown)


Sreekant13:
Rebased onto main; the only conflict was a changelog entry landing in the same spot, now resolved with both kept.

Sreekant13:
A gentle check-in on this one, and no rush.

`StrReplaceFile` counted `total_replacements` against the original file content rather than the running content, so chained edits under-report: replacing `hello` with `goodbye` and then `goodbye` with `farewell` reports 1 when 2 replacements happened. The fix counts against the running content, with a regression test and a changelog entry, and it resolves #2526.

It was rebased onto main in August, so the changelog conflict is long gone and it is mergeable. Happy to rebase again if it has drifted since.

PR #2530, opened 2026-07-21
fix(shell): stop blocking until timeout when a detached child holds the pipes

### LINKED ISSUE #2468: Kimi CLI hangs after detached child-process tool call
## What version of Kimi Code CLI is running?

Kimi CLI `1.47.0`.

## Which open platform/subscription were you using?

API-compatible local mock provider with a fake API key.

## Which model were you using?

OpenAI-compatible local mock provider.

## What platform is your computer?

Linux x86_64 Docker.

## What issue are you seeing?

When Kimi CLI receives a shell tool call that starts a detached child process, the CLI does not return control within the bounded run. The standalone reproducer downloads public `Kimi CLI 1.47.0`, starts a local OpenAI-compatible mock provider, sends a child-spawning shell tool call, and verifies the behavior without requiring this repository or any private fuzzing harness.

Kimi CLI accepts the child-spawning shell tool call, records the detached child PID side effect, and then fails to finish normally before the bounded timeout. The reproducer prints `REPRODUCED` only when provider traffic is observed, the child-spawning command records its PID, and the CLI exits by timeout or signal.

## What steps can reproduce the bug?

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It downloads the affected public Linux release asset, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/707db65552981ab1c84109517315bd2f

The Gist contains `kimi-cli-child-process-cleanup.reproduce.py`. Download and run:

```bash
gh gist clone 707db65552981ab1c84109517315bd2f kimi-cli-child-process-cleanup-reproducer
cd kimi-cli-child-process-cleanup-reproducer
python3 kimi-cli-child-process-cleanup.reproduce.py
```

To reuse an already-built local image:

```bash
python3 kimi-cli-child-process-cleanup.reproduce.py --skip-build
```

Key output from a local run against the affected version:

```text
provider_requests_at_least: observed=... expected>=1
side_effect_file_exists detached-child.pid: bytes=... expected>=1
timeout_or_signal: returncode=... timed_out=True
REPRODUCED
```

## What is the expected behavior?

Kimi CLI should track, terminate, and reap tool-started child processes, then return control without hanging.

## Additional information

- Target: `MoonshotAI/kimi-cli`
- Affected version checked with the public reproducer: `1.47.0`
- Current public latest checked on 2026-06-22: GitHub release `1.47.0`
- Environment: Linux x86_64 Docker, public release asset
- Provider/model: local OpenAI-compatible mock provider, `gpt-4o`

This report targets stock public `Kimi CLI 1.47.0` from the GitHub release asset. No upstream issue has been filed from this draft yet.

ayaangazali:
## Related Issue

Resolve #2468

## Description

In the foreground shell path, `_run_shell_command` waits for stdout/stderr EOF (`asyncio.wait_for(gather(read stdout, read stderr), timeout)`) before ever checking the exit code. A command like `some_daemon & echo done` leaves a detached child holding the write ends of the pipes, so EOF never arrives: the tool blocks for the entire command timeout even though the shell exited immediately, then reports "Command killed by timeout" for a command that actually succeeded. That is the hang in #2468.

Two things I learned while fixing it, which shaped the implementation:

1. Waiting on `process.wait()` instead does not work: asyncio gates its subprocess exit waiters on pipe disconnection, so `wait()` blocks exactly the same way (verified empirically on macOS and Linux Python 3.14: `wait()` returned only when the detached child died, while `returncode` was set within 50ms of the shell exiting). So exit is observed by polling `returncode` at 50ms intervals, which the `KaosProcess` protocol guarantees on every backend.
2. Abandoning the reader `gather` after cancel triggers asyncio's "_GatheringFuture exception was never retrieved" loop-level error, which prompt_toolkit surfaces as a full-screen "Unhandled exception in event loop" banner that freezes the TUI. The existing PTY e2e tests (`test_ctrl_c_during_running_turn_interrupts`, `test_shell_cancel_running_command_kills_process_and_recovers`) caught this in my first attempt, hence the `_consume_exception` done callback.

Behavior after the fix:

- Shell exits while something still holds the pipes: remaining output is drained for a bounded `PIPE_DRAIN_GRACE` (2s), then the tool returns the real exit code. Output produced by a detached child after that grace is dropped, which I think is the right semantic: the shell finishing is what "the command finished" means.
- Genuinely long-running commands: unchanged, still killed at the timeout with the same error message.
- Ctrl-C cancel: unchanged (kill + raise), minus the event loop leak.
- Normal fast commands: unchanged output, worst case ~50ms added exit-detection latency.

Note this intentionally does not implement full child-process tracking/reaping that the issue mentions as the ideal. The detached child keeps running (same as before the fix); the change just stops it from stalling the agent loop. Figured the bigger cleanup story needs maintainer direction first.

Tests: new `test_detached_child_holding_pipes_does_not_block_until_timeout` (fails at 25s timeout before the fix, returns in ~2s after). The fake processes in existing tests grew a `returncode` attribute since the code now polls it, and the cancel-test fake now models "running until killed" to keep testing the same scenario it did before.

## Checklist

- [x] I have run `make gen-changelog` to update the changelog. (hand-edited CHANGELOG.md in the same style, i don't have the Kimi API setup the skill needs)
- [x] I have run `make gen-docs` to update the user documentation. (no user-facing docs describe the shell tool's pipe/timeout behavior, nothing to regenerate)

---

being upfront: freshman here, i dug through this one with claude code as a rubber duck for the asyncio semantics (the wait() thing surprised both of us). tested everything i could locally but if the approach is off i'm all ears, learning as i go :)

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2530" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Only write entries that are worth mentioning to users.
 ## Unreleased
 
 - Kosong: Stop sending an empty `anthropic-beta` header when no beta features are declared — adaptive thinking removes the interleaved-thinking beta, which previously left an empty header value that some backends reject
+- Shell: Fix the shell tool blocking until the full command timeout when a detached child process inherits stdout/stderr, then wrongly reporting a timeout kill. The tool now returns shortly after the shell itself exits and drains remaining pipe output for a bounded grace period
 
 ## 1.49.0 (2026-07-16)
 
--- a/src/kimi_cli/tools/shell/__init__.py
+++ b/src/kimi_cli/tools/shell/__init__.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 from collections.abc import Callable
 from pathlib import Path
 from typing import Self, override
@@ -21,6 +22,18 @@ from kimi_cli.utils.subprocess_env import get_noninteractive_env
 
 MAX_FOREGROUND_TIMEOUT = 5 * 60
 MAX_BACKGROUND_TIMEOUT = 24 * 60 * 60
+PIPE_DRAIN_GRACE = 2.0
+"""Seconds to keep draining stdout/stderr after the shell process exits.
+
+A detached child that inherited the pipes can keep them open long after the
+shell itself has exited; without a bound the tool would block until the full
+command timeout waiting for an EOF that may never come."""
+EXIT_POLL_INTERVAL = 0.05
+"""Seconds between exit checks while the shell process is running.
+
+`KaosProcess.wait()` may not resolve until all pipes close (asyncio gates its
+exit waiters on pipe disconnection), so process exit is observed by polling
+`returncode` instead."""
 
 
 class Params(BaseModel):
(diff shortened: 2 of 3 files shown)


ayaangazali:
Quick status on this one since it has been open a couple of weeks.

What it fixes: the foreground shell path waited for stdout/stderr EOF before checking the exit code, so a command that leaves a detached child holding the pipes (`some_daemon & echo done`) blocked for the whole timeout and then got reported as killed by timeout even though it succeeded. Resolves #2468.

Branch is still current with main (main has not moved since 4a550eff, so there is nothing to rebase onto) and it still merges clean. `make check-kimi-cli` and `make test-kimi-cli` both passed locally, including the two PTY e2e cancel tests.

One thing worth flagging: the CI checks here have never actually run, they are all sitting at `action_required`, which I believe is the approval gate for first time contributors. So the empty check list is not a failure, it is just waiting on someone to approve the run.

Happy to adjust the approach or drop it if it is not the direction you want.

PR #2539, opened 2026-07-23
fix(mcp): normalize tools for Moonshot API

### LINKED ISSUE #2531: MCP tool names & schemas rejected by Moonshot API (HTTP 400) — sanitize client-side before sending
**Environment:** kimi-cli 1.49.0 · macOS (arm64) · Model: K3 via Moonshot API

The API rejects such requests with:

```
Error: [provider.api_error] 400 tools.function.parameters is not a valid
moonshot flavored json schema, details: <At path 'root': when using anyOf,
type should be defined in anyOf items instead of the parent schema>
```

## Summary

Several spec-valid MCP servers cannot be used with kimi-cli, because the Moonshot
chat-completions endpoint rejects their tool definitions with **HTTP 400**. The MCP
spec itself allows these names/schemas — other API consumers handle them fine.
kimi-cli (or the API) should normalize tool definitions before the request.

## Problem 1: Tool names starting with a digit → HTTP 400

The MCP server `@21st-dev/magic` (a popular UI-component generator) exposes tools
named:

```
21st_magic_component_builder
21st_magic_component_inspiration
21st_magic_component_refiner
...
```

MCP places no constraint on the leading character of tool names. The Moonshot API,
however, rejects any request whose `tools[].function.name` begins with a digit:

- Same session, same prompt, same tools — the call fails with HTTP 400 as long as
  the names start with `21st_…`.
- Renaming the **identical** tools to `m_21st_magic_*` (letter first) makes the
  exact same calls succeed. No other change.

**Ask:** Either relax the API-side pattern (e.g. OpenAI-compatible
`^[a-zA-Z0-9_-]{1,64}$`, which permits leading digits), or have kimi-cli rewrite
non-conforming tool names client-side (with a stable mapping so results route back
to the correct MCP tool).

## Problem 2: Strict JSON-Schema subset rejects valid MCP schemas

Example: `gitnexus@1.6.9` (`npx gitnexus mcp`), tool `api_impact`, declares this
inputSchema — valid JSON Schema (2020-12):

```json
{
  "type": "object",
  "properties": {
    "route":  { "type": "string", "description": "Route path (e.g., \"/api/grants\")" },
    "file":   { "type": "string", "description": "Handler file path (alternative to route)" },
    "method": { "type": "string" },
    "repo":   { "type": "string" },
    "branch": { "type": "string" }
  },
  "required": [],
  "anyOf": [
    { "required": ["route"] },
    { "required": ["file"] }
  ]
}
```

The API rejects it (HTTP 400). The offending constructs appear to be the root-level
`anyOf` with per-branch `required`, combined with the empty sibling `"required": []`.
Rewriting the schema to drop the empty `required` and resolve the `anyOf` (moving
the constraint into the description) makes the identical tool call succeed.

Related: some servers omit the root `"type": "object"` entirely; coercing it
client-side is trivial and harmless.

(For completeness: GitNexus fixed the `anyOf` part upstream in 1.6.10-rc via
PR #2489 — verified by auditing all 17 tool schemas on `gitnexus@rc`. The
client-side problem for kimi-cli remains: any spec-valid server can still hit
these rejections, e.g. digit-led tool names or other combinator patterns.)

## Suggested fix (client-side, minimal)

Before sending `tools` to the API, kimi-cli could:

1. Rewrite names to `^[A-Za-z][A-Za-z0-9_-]*$` (e.g. prefix `m_` for leading
   digits, replace invalid chars with `_`), keeping a bidirectional map for the
   response path.
2. Recursively coerce schemas: force root `type: "object"` when `properties` is
   present; when `anyOf`/`oneOf` appears alongside sibling keywords, either inline
   the siblings into each branch or drop the conflicting sibling with a note in
   `description`.

This is exactly what our local workaround does (a FastMCP stdio proxy between
kimi-cli and the upstream server). It works, but every user of these servers
shouldn't need to build their own proxy — the sanitization belongs in the client
(or the API validation should be relaxed).

## Repro

```bash
# 1. magic (name problem)
kimi mcp add magic -- npx -y @21st-dev/magic@latest
# ask anything that loads tools -> HTTP 400

# 2. gitnexus (schema problem)
kimi mcp add gitnexus -- npx -y gitnexus@latest mcp
# ask anything that loads tools -> HTTP 400
```

Happy to provide the proxy source or test a fix.

lihailong00:
## Summary

- generate stable Moonshot-compatible aliases for MCP tool names while retaining original names for upstream call routing
- add a missing root `object` type when an MCP schema defines object properties
- distribute the exact object `anyOf`/required schema shape reported in the issue into equivalent Moonshot-compatible branches
- leave unsupported complex combinator schemas unchanged

## Validation

- `uv run pytest packages/kosong/tests/test_kimi_tool_conversion.py tests/core/test_mcp_tool_names.py -q` (9 passed)
- targeted Ruff check and format check passed
- targeted Pyright check passed with 0 errors
- remote compare verified: 1 commit, 4 intended files, 154 additions / 2 deletions

Closes #2531

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2539" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/kosong/src/kosong/chat_provider/kimi.py
+++ b/packages/kosong/src/kosong/chat_provider/kimi.py
@@ -44,6 +44,7 @@ from kosong.message import (
 )
 from kosong.tooling import Tool
 from kosong.utils.jsonschema import JsonDict, ensure_property_types
+from kosong.utils.typing import JsonType
 
 if TYPE_CHECKING:
 
@@ -383,9 +384,55 @@ def _convert_tool(tool: Tool) -> ChatCompletionToolParam:
     function = converted["function"]
     parameters = function.get("parameters")
     if isinstance(parameters, dict):
-        normalized = ensure_property_types(cast(JsonDict, parameters))
+        normalized = copy.deepcopy(cast(JsonDict, parameters))
+        if isinstance(normalized.get("properties"), dict):
+            normalized.setdefault("type", "object")
+        normalized = _normalize_object_any_of(normalized)
+        normalized = ensure_property_types(normalized)
         function["parameters"] = cast(dict[str, object], normalized)
     return converted
+
+
+def _normalize_object_any_of(schema: JsonDict) -> JsonDict:
+    """Distribute a simple object schema into ``anyOf`` required branches.
+
+    Moonshot rejects ``type`` beside ``anyOf``. MCP tools commonly use this
+    shape solely to require one of several object properties. Distributing the
+    shared object constraints into those branches preserves the schema's
+    meaning while satisfying Moonshot's stricter dialect.
+    """
+    branches = schema.get("anyOf")
+    if not (
+        schema.get("type") == "object"
+        and isinstance(schema.get("properties"), dict)
+        and isinstance(branches, list)
+        and branches
+        and all(
+            isinstance(branch, dict)
+            and set(branch).issubset({"required"})
+            and isinstance(branch.get("required", []), list)
+            for branch in branches
+        )
+    ):
+        return schema
+
+    common = copy.deepcopy(schema)
+    common.pop("anyOf")
+    common_required = common.pop("required", [])
+    if not isinstance(common_required, list):
+        return schema
+
+    distributed: list[JsonType] = []
+    for branch in branches:
+        assert isinstance(branch, dict)
+        branch_required = branch.get("required", [])
+        assert isinstance(branch_required, list)
+        branch_schema = copy.deepcopy(common)
+        required = list(dict.fromkeys([*common_required, *branch_required]))
+        if required:
+            branch_schema["required"] = required
+        distributed.append(branch_schema)
+    return {"anyOf": distributed}
 
 
 class KimiStreamedMessage:
--- a/packages/kosong/tests/test_kimi_tool_conversion.py
+++ b/packages/kosong/tests/test_kimi_tool_conversion.py
@@ -81,3 +81,55 @@ def test_convert_tool_passes_through_already_typed_schema() -> None:
         "properties": {"msg": {"type": "string"}},
         "required": ["msg"],
     }
+
+
+def test_convert_tool_distributes_object_constraints_into_any_of() -> None:
+    tool = Tool(
+        name="api_impact",
+        description="Find an API route or file.",
+        parameters={
+            "type": "object",
+            "properties": {
+                "route": {"type": "string"},
+                "file": {"type": "string"},
+            },
+            "required": [],
+            "anyOf": [{"required": ["route"]}, {"required": ["file"]}],
+        },
+    )
+
+    parameters = _convert_tool(tool)["function"].get("parameters")
+
+    assert parameters == {
+        "anyOf": [
+            {
+                "type": "object",
+                "properties": {
+                    "route": {"type": "string"},
+                    "file": {"type": "string"},
+                },
+                "required": ["route"],
+            },
+            {
+                "type": "object",
+                "properties": {
+                    "route": {"type": "string"},
+                    "file": {"type": "string"},
+                },
+                "required": ["file"],
+            },
+        ]
+    }
+
+
+def test_convert_tool_adds_missing_root_object_type() -> None:
+    tool = Tool(
+        name="echo",
+        description="Echo input.",
+        parameters={"properties": {"message": {"type": "string"}}},
+    )
+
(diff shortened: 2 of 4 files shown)


zycxfyh:
Thanks for addressing #2531. I checked out `0861fd3` and independently validated the patch.

### Validation that passed

- `uv run pytest packages/kosong/tests/test_kimi_tool_conversion.py tests/core/test_mcp_tool_names.py -q` → **9 passed**
- targeted Ruff → **passed**
- targeted Pyright → **0 errors**
- 98,980 generated invalid `(server_name, tool_name)` pairs → **0 collisions among generated aliases**
- the `anyOf` transformation preserved shared `required`, `title`, `description`, `properties`, and `additionalProperties` in supplemental checks

The reported `21st_*` and simple object-`anyOf` cases look correctly handled.

### One routing collision remains

`_safe_mcp_tool_name()` cannot guarantee uniqueness against an already-valid upstream name, because compatible names are returned unchanged. A server (or another loaded server) can expose a valid tool whose original name is exactly the alias generated for an invalid tool:

```python
server = "magic"
invalid = "21st.magic component builder"
alias = _safe_mcp_tool_name(server, invalid)
# alias == "m_21st_magic_component_builder_25ae5ad3"

_safe_mcp_tool_name(server, alias) == alias
```

Constructing both `MCPTool`s gives the same runtime name:

```text
runtime_a m_21st_magic_component_builder_25ae5ad3
runtime_b m_21st_magic_component_builder_25ae5ad3
same_runtime_name True
```

`KimiToolset.add()` currently performs:

```python
self._tool_dict[tool.name] = tool
```

so registration silently overwrites the first tool:

```text
registered_count 1
winner_upstream m_21st_magic_component_builder_25ae5ad3
```

The model-visible call can therefore be routed to a different original MCP tool depending on load order.

### Suggested fix

Resolve aliases with visibility over the complete tool set rather than independently, or at minimum detect duplicate runtime names during MCP registration and fail with an explicit diagnostic instead of silently replacing an existing tool. A regression test pairing an invalid name with a valid name equal to its generated alias should cover the case.

This does not invalidate the schema normalization or the ordinary #2531 examples, but I think the silent routing ambiguity is worth closing before relying on the alias map as bidirectional.

zycxfyh:
Here is the compact regression test corresponding to the collision report. It can be copied into `tests/core/test_mcp_alias_collision_regression.py`:

```python
from types import SimpleNamespace
from typing import Any, cast

import pytest
from mcp.types import Tool

from kimi_cli.soul.toolset import KimiToolset, MCPTool, _safe_mcp_tool_name


def test_mcp_alias_collision_is_rejected_instead_of_silent_overwrite() -> None:
    server_name = "magic"
    invalid_name = "21st.magic component builder"
    generated_alias = _safe_mcp_tool_name(server_name, invalid_name)

    # A different MCP tool is allowed to use that generated alias as its
    # original, already-compatible name.
    assert _safe_mcp_tool_name(server_name, generated_alias) == generated_alias

    runtime = SimpleNamespace(
        config=SimpleNamespace(
            mcp=SimpleNamespace(client=SimpleNamespace(tool_call_timeout_ms=60_000))
        )
    )
    client = cast(Any, object())

    invalid_tool = MCPTool(
        server_name,
        Tool(name=invalid_name, description="invalid original", inputSchema={"type": "object"}),
        client,
        runtime=cast(Any, runtime),
    )
    valid_tool = MCPTool(
        server_name,
        Tool(
            name=generated_alias,
            description="valid original",
            inputSchema={"type": "object"},
        ),
        client,
        runtime=cast(Any, runtime),
    )

    assert invalid_tool.name == valid_tool.name

    toolset = KimiToolset()
    toolset.add(invalid_tool)

    # Minimal safe policy: reject the duplicate explicitly. An implementation
    # that allocates a second unique runtime alias can use an equivalent
    # assertion that both original tools remain registered and routable.
    with pytest.raises(ValueError, match="duplicate|conflict|already registered"):
        toolset.add(valid_tool)
```

Run with:

```bash
uv run pytest tests/core/test_mcp_alias_collision_regression.py -q
```

Against commit `0861fd3`, it currently fails at the intended single point:

```text
Failed: DID NOT RAISE <class 'ValueError'>
```

This version encodes the minimum fail-safe policy: duplicate runtime names must be rejected rather than silently overwrite an existing MCP route. If the implementation instead allocates a second unique alias, the final assertion can be changed to require both original tools to remain registered and independently routable.

PR #2554, opened 2026-07-23
fix(tools): count StrReplaceFile replacements against running content

ayaangazali:
## Related Issue

No open issue for this one, I hit it while reading `StrReplaceFile`. It is a small self-contained correctness fix (well under the 100 LOC discussion threshold), happy to close if you would rather have an issue first.

## Description

The success message for `StrReplaceFile` counted replacements by summing `original_content.count(edit.old)` for each edit. But edits apply sequentially, so a later edit runs against the output of the earlier ones, not the original file. When edits interact, the reported total is wrong.

Example: file `foo bar foo`, two `replace_all` edits `foo -> "foo bar"` then `bar -> "baz"`:
- `foo -> "foo bar"` replaces 2, giving `foo bar bar foo bar`
- `bar -> "baz"` then replaces 3, giving `foo baz baz foo baz`
- actual total 5, but the message reported 3 (2 + 1 counted against the original)

The file edit itself was always correct; only the reported count was off. Fix counts each replacement as it actually happens by having `_apply_edit` return the count alongside the new content, which also lets the redundant second pass over the edits go away.

Single-edit cases are unchanged (a `replace_all` edit still reports its occurrence count, a single edit reports 1).

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works. (two tests in `test_str_replace_file.py`: a single-edit count check and the interacting-edits case above, which reported 3 before)
- [x] I have run `make gen-changelog` to update the changelog. (hand-edited CHANGELOG.md in the same style, I do not have the Kimi API setup the skill needs)
- [x] I have run `make gen-docs` to update the user documentation. (no user docs describe this message, nothing to regenerate)

---

quick note: freshman here, found this reading the edit tool and figured a wrong count could quietly mislead. kept the change tiny and consulted claude code on the counting semantics to be sure str.count lines up with str.replace. if you would rather not touch it, no worries at all, still learning what is worth a PR :)

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2554" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Only write entries that are worth mentioning to users.
 ## Unreleased
 
 - Kosong: Stop sending an empty `anthropic-beta` header when no beta features are declared — adaptive thinking removes the interleaved-thinking beta, which previously left an empty header value that some backends reject
+- Tools: Fix the StrReplaceFile success message reporting the wrong replacement total when multiple edits interact; the count now follows the running content instead of the original file
 
 ## 1.49.0 (2026-07-16)
 
--- a/src/kimi_cli/tools/file/replace.py
+++ b/src/kimi_cli/tools/file/replace.py
@@ -78,12 +78,15 @@ class StrReplaceFile(CallableTool2[Params]):
             )
         return None
 
-    def _apply_edit(self, content: str, edit: Edit) -> str:
-        """Apply a single edit to the content."""
+    def _apply_edit(self, content: str, edit: Edit) -> tuple[str, int]:
+        """Apply a single edit, returning the new content and how many
+        replacements it made against the content passed in."""
         if edit.replace_all:
-            return content.replace(edit.old, edit.new)
+            count = content.count(edit.old)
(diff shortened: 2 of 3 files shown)


ayaangazali:
Status update on this one.

What it fixes: `StrReplaceFile` counts replacements against the original file content while the edits actually apply sequentially, so when a later edit matches text an earlier edit produced, the reported total is wrong. The file is written correctly, only the number in the result message is off.

Worth noting #2526 was opened after this PR by someone who hit the same thing independently, so it is not purely theoretical. Their repro (`hello -> goodbye -> farewell` reporting 1 instead of 2) is a cleaner example than mine.

Branch is current with main (main has not moved since 4a550eff, nothing to rebase onto) and still merges clean. `make check-kimi-cli` and `make test-kimi-cli` passed locally.

Same note as my other PRs: the checks here show as `action_required` rather than passing or failing, which I think is the first time contributor approval gate, so they have not run yet.

Glad to change the approach if you would rather solve it a different way.

PR #2572, opened 2026-07-31
fix(kosong): recursively unwrap double-encoded JSON in tool-call arguments

### LINKED ISSUE #2406: bug: Tool call arguments double-encoding breaks array/dict parameters (Moonshot API)
## Problem

Tool calls with array or object parameters (e.g. `SetTodoList`, `ExitPlanMode`, `StrReplaceFile`) fail with Pydantic validation errors:

```
Error validating JSON arguments: 1 validation error for Params
todos
  Input should be a valid list [type=list_type, input_value=[{"title": ...}], input_type=str]
```

The `todos` parameter is received as a **JSON string** instead of a parsed list/array.

## Root Cause

In `KimiToolset.handle()` (`kimi_cli/soul/toolset.py:301`):

```python
arguments: JsonType = json.loads(tool_call.function.arguments or "{}", strict=False)
```

The Moonshot API returns `function.arguments` with **double-encoded JSON strings** for nested array/object values. After `json.loads` parses the outer JSON, inner values like `todos` remain as strings:

```
API returns: {"todos": "[{\"title\": \"test\", \"status\": \"in_progress\"}]"}
                              ^--- string, not array!
json.loads → {"todos": "[{\"title\": \"test\", \"status\": \"in_progress\"}]"}
                        ^--- still a string!
Pydantic → ERROR: expected list, got str
```

## Affected Tools

| Tool | Parameter | Type |
|------|-----------|------|
| `SetTodoList` | `todos` | `list[Todo]` |
| `ExitPlanMode` | `options` | `list[Option]` |
| `StrReplaceFile` | `edit` | `dict` or `list[Edit]` |
| Any tool with array/dict params | varies | varies |

## Reproduction

1. Call `SetTodoList` with `todos=[{"title": "test", "status": "in_progress"}]`
2. Observe Pydantic validation error

## Proposed Fix

In `kimi_cli/soul/toolset.py`, `KimiToolset.handle()` method, replace:

```python
arguments: JsonType = json.loads(tool_call.function.arguments or "{}", strict=False)
```

with:

```python
raw = tool_call.function.arguments or "{}"
if isinstance(raw, dict):
    arguments: JsonType = raw
else:
    arguments = json.loads(raw, strict=False)
    # Moonshot API quirk: double-encoded JSON strings
    if isinstance(arguments, dict):
        for k, v in list(arguments.items()):
            if isinstance(v, str) and v and v[0] in ("[", "{"):
                try:
                    arguments[k] = json.loads(v, strict=False)
                except (json.JSONDecodeError, ValueError):
                    pass
```

## Environment

- Kimi CLI: 1.46.0
- Pydantic: 2.12.5
- OpenAI SDK: 2.14.0
- Python: 3.14
- OS: Linux

## Additional Context

The same issue exists in `SimpleToolset.handle()` in `kosong/tooling/simple.py` (the upstream library), but the Kimi CLI uses `KimiToolset` instead, which has its own `handle()` method with the same bug.

aalhadxx:
## Problem

Tool calls with array or object parameters (e.g. SetTodoList, ExitPlanMode, StrReplaceFile) fail with Pydantic validation errors when using providers that double-encode nested values:

The Moonshot API returns "function.arguments" where inner array/object values are themselves JSON strings. A single json.loads leaves these as strings, which then fail validation.

A prior fix was attempted in #2513 (Devin found 1 potential issue; author has not responded). This PR supersedes it with a cleaner implementation and broader test coverage.

## Solution

Introduce  — a shared helper consumed by both  and .

Key improvements over #2513:
- **Fast-path prefix check**: strings are only parsed if their first non-whitespace char is  or , avoiding unnecessary  calls on typical non-JSON string values.
- **Outer malformed guard**:  is re-raised for malformed outer input so callers can surface  (preserves today's contract).
- **Scalar protection**: strings that parse to scalars (, , ) are left untouched so genuine string fields are never corrupted.
- **Recursion terminates**:  only re-enters when a string parses to a dict/list; non-string leaves return immediately.

## Files changed

| File | Change |
|------|--------|
|  | New shared helper |
|  | Replace json.loads with helper |
|  | Replace json.loads with helper |
|  | 23 unit tests (E1–E10, termination, regression, unicode) |
|  | 3 integration tests through KimiToolset |

## Tests

Unit tests cover:
- E1–E2: top-level and nested double-encoding unwrap
- E3: non-JSON strings preserved (malformed braces, empty strings)
- E4: scalar JSON strings preserved (int, float, bool, null)
- E5: None / empty input coerces to 
- E6: well-formed single-encoded args unchanged (regression)
- E7: list-typed outer value with inner string decoded
- E8: mid-bracket strings and leading-whitespace handling
- E9: malformed outer input re-raises JSONDecodeError
- E10: fast-path sanity — plain-text strings never parsed
- Termination: deep nested arrays (20 levels) and double-encoded chains (10 levels)
- Regression: SetTodoList, StrReplaceFile.edit, ExitPlanMode.options shapes
- Mixed genuine + encoded fields in the same dict
- Unicode in double-encoded strings

Integration tests verify the fix end-to-end through  with a dummy tool mirroring 's  parameter.

Fixes #2406
Supersedes / improves upon #2513

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2572" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


aalhadxx:
@devin-ai-integration good catch — you are absolutely right. Unconditionally unwrapping every string is too aggressive.

Pushed a fix that makes the unwrapping **failure-driven**:

1.  now tries strict  first (no recursive unwrapping)
2. Calls the tool with those arguments
3. If validation fails (), it retries with  (recursive unwrapping)
4. Returns whichever attempt succeeds

This means genuine JSON text in  or  strings stays untouched unless the strict parse actually fails validation — which is exactly the case for double-encoded Moonshot API responses.

Also added regression tests documenting the  case. Let me know if you see any other issues.

PR #2577, opened 2026-08-01
fix(web,vis): do not crash printing the startup banner on legacy console codecs

### LINKED ISSUE #2532: kimi web crashes at startup on Windows when stdout is redirected: UnicodeEncodeError (gbk) in print_banner
## Bug description

`kimi web` crashes immediately at startup on Chinese-locale Windows when its stdout is redirected (pipe/file) or when it is launched by a parent process that captures stdout. The crash happens while printing the startup banner: the `➜` (U+279C) character cannot be encoded with the locale codec (GBK / cp936), and the unhandled `UnicodeEncodeError` kills the whole process **before the HTTP server binds its port**.

As a result, the web server never starts, and a Web UI page already open in the browser shows repeated "实时连接出错 / WebSocket error" toasts because nothing is listening.

## Environment

- OS: Windows 11 (10.0.26100), Simplified Chinese locale (ANSI codepage 936 / GBK)
- Kimi Code CLI: v0.28.1, native win32-x64 build (bundled with VS Code extension `moonshot-ai.kimi-code`, `kimi_cli` 1.43.0, PyInstaller / Python 3.14)

## Steps to reproduce

1. On a Chinese-locale Windows machine, redirect stdout, e.g.:
   ```
   kimi.exe web --port 58627 --no-open > web.log 2>&1
   ```
   (or launch `kimi web` from any parent process that pipes stdout)
2. The process exits immediately; the port is never bound.

## Actual behavior

```
Traceback (most recent call last):
  ...
  in web:70
  in run_web_server:437
  in print_banner:118
UnicodeEncodeError: 'gbk' codec can't encode character '➜' in position 4:
illegal multibyte sequence
[PYI-xxxxx:ERROR] Failed to execute script '__main__' due to unhandled exception!
```

## Expected behavior

Banner printing should never crash the server. Please make `print_banner` encoding-safe (e.g. write with `errors="replace"`, force UTF-8, or drop non-encodable characters) so `kimi web` starts regardless of stdout encoding / console availability.

## What does NOT help

`chcp 65001`, `PYTHONUTF8=1`, and `PYTHONIOENCODING=utf-8` all fail to prevent the crash when stdout is redirected (verified on v0.28.1).

## Workaround

Run `kimi web` attached to a real console window (cmd / PowerShell / Windows Terminal) and keep that window open — console I/O uses UTF-8 (PEP 528), so the banner prints fine and the server starts normally.

ayaangazali:
## Related Issue

Resolve #2532

## Description

`print_banner` in `src/kimi_cli/utils/server.py` writes the banner with a bare `print()`, and both callers put U+279C in front of each URL (`web/app.py:374,376` and `vis/app.py:128`). On a console whose codec cannot represent that character, GBK on Chinese Windows being the reported case, the print raises `UnicodeEncodeError`.

The part that makes it more than cosmetic: the banner is printed before the server binds its port, so the unhandled error takes the process down and nothing ends up listening. From the user side that shows up as a Web UI stuck on WebSocket errors rather than as a crash, which makes it a confusing one to diagnose.

Characters the current stdout encoding cannot represent are now replaced instead of raising. Two details worth mentioning:

- The substitution happens before the box width is measured, otherwise the replacement changes the line length and the box borders stop lining up.
- Terminals that can encode the glyph are unaffected, so this does not downgrade normal output. There is a test covering that.

I deliberately did not touch `sys.stdout` reconfiguration or add an env var. Both would fix the symptom globally but change behavior well outside this banner, and the crash only needs the one guard where both callers already funnel through.

The issue was filed against 1.43.0; I verified the same code path is still there at HEAD (4a550eff, 1.49.0).

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works. (renders the banner through a gbk stream, which raises on current main; plus a box alignment check and a utf-8 case asserting the arrow survives)
- [x] I have run `make gen-changelog` to update the changelog. (hand-edited CHANGELOG.md in the same style, I do not have the Kimi API setup the skill needs)
- [x] I have run `make gen-docs` to update the user documentation. (nothing in docs covers banner rendering)

Note on gates: `make check-kimi-cli` is clean and `make test-kimi-cli` passes. One PTY e2e test (`test_shell_mode_toggle_roundtrip`) flaked once under full suite load on my machine; it passes in isolation on both this branch and unmodified main, and the full e2e suite passes on re-run, so it looks like timing rather than anything from this change.

---

small disclosure as usual: freshman, no Windows machine to hand so I reproduced the encoding failure by rendering the banner through a gbk stream rather than on real hardware. logic decisions talked through with claude code. if you would rather just swap the arrow for an ascii character and skip the helper entirely, that is a totally fair call and i am happy to redo it that way :)

---

<a href="https://app.devin.ai/review/moonshotai/kimi-cli/pull/2577" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Only write entries that are worth mentioning to users.
 ## Unreleased
 
 - Kosong: Stop sending an empty `anthropic-beta` header when no beta features are declared — adaptive thinking removes the interleaved-thinking beta, which previously left an empty header value that some backends reject
+- Web/Vis: Fix `kimi web` and `kimi vis` dying at startup on consoles whose codec cannot encode the banner arrow (GBK on Chinese Windows, for example). The banner is printed before the server binds its port, so the crash left nothing listening; unsupported characters are now replaced instead of raising
 
 ## 1.49.0 (2026-07-16)
 
--- a/src/kimi_cli/utils/server.py
+++ b/src/kimi_cli/utils/server.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import importlib
 import socket
+import sys
 import textwrap
 
 
@@ -86,8 +87,28 @@ def get_network_addresses() -> list[str]:
     return addresses
 
 
+def _encodable(text: str) -> str:
+    """Drop characters stdout cannot encode, so printing never raises.
+
+    The banner uses characters like U+279C that a legacy console codec (GBK on
+    Chinese Windows, for example) cannot represent. Printing those raises
+    UnicodeEncodeError, and since the banner is printed before the server binds
+    its port, an unhandled error there takes the whole process down.
+    """
+    encoding = getattr(sys.stdout, "encoding", None) or "utf-8"
+    try:
+        text.encode(encoding)
+    except UnicodeEncodeError:
+        return text.encode(encoding, errors="replace").decode(encoding, errors="replace")
+    except LookupError:
+        return text
+    return text
+
+
 def print_banner(lines: list[str]) -> None:
     """Print a boxed banner with tag conventions (<center>, <nowrap>, <hr>)."""
+    # Sanitize before measuring so the box borders still line up.
+    lines = [_encodable(line) for line in lines]
     processed: list[str] = []
     for line in lines:
         if line == "<hr>":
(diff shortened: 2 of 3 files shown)


ayaangazali:
Rebased onto cbc15c07. The only conflict was the CHANGELOG `## Unreleased` entry landing in the same spot as the kosong anthropic-beta fix, so both entries are kept now. No code changes from the rebase.

Re-ran the gates after rebasing: `make check-kimi-cli` clean, `make test-kimi-cli` 2959 passed plus the e2e suite. Merges clean again.

Issue #2191, opened 2026-05-08
[Windows] StrReplaceFile silently converts entire file from CRLF to LF, forcing Agent to abandon native tools for Python workarounds


lNeverl:
**Related Issue**: #1952 (CRLF issue with StrReplaceFile)  
**Related PR**: #1953 (fix(file): preserve CRLF during file edits) -- still open/unmerged

---

#### Problem Description

On Windows, repository files use standard `CRLF` (\r\n) line endings (checked out via Git `core.autocrlf=true`). When the Agent attempts to use `StrReplaceFile` for localized edits, it hits a **double failure**:

1. **Using \n (LF) in the `old` string**: The match succeeds, but `StrReplaceFile` **rewrites the entire file from CRLF to LF**, producing a massive unrelated diff.
2. **Using \r\n (CRLF) in the `old` string**: The match fails outright with `"old string was not found"`.

This makes `StrReplaceFile` **completely unsafe to use** on Windows CRLF files. The Agent is forced to fall back to executing Python scripts through the `Shell` tool for every file modification.

---

#### Cascading Consequences (Real Session)

Once forced into Python-via-Shell on Windows PowerShell, a chain of secondary failures occurs:

| Failure | Symptom | Frequency |
|---------|---------|-----------|
| `SyntaxError: EOF while scanning triple-quoted string literal` | Multi-line Python passed via `python -c "..."` gets truncated | Multiple |
| `SyntaxError: EOL while scanning string literal` | Backslash/ newline escaping across PowerShell + Python layers fails | Multiple |
| `SyntaxError: f-string expression part cannot include a backslash` | Windows path backslashes inside f-strings cause syntax errors | Multiple |
| `AttributeError: 'bytes' object has no attribute 'encode'` | Double `.encode()` on a bytes object in the script | Once |
| JSON parse error (`Unterminated string`) | Unescaped quotes inside `SetTodoList` / tool-call JSON payloads | Multiple |
| Tab vs. space mismatch | Some source files mix \t and space indentation; Python scripts searching with the wrong whitespace type fail to find targets | Multiple |

**End result**: Modifying ~7 source files consumed **110 tool calls across 4 conversation turns**, and the files had to be redone from scratch after an accidental revert.

---

#### Reproduction

**Repro 1 -- \n old string silently corrupts line endings**

1. Take any standard Windows CRLF source file (e.g., a C header with 300+ lines, all \r\n).
[...]

PR #2595, opened 2026-08-06
fix(StrReplaceFile): refuse to edit files that are not valid UTF-8

### LINKED ISSUE #2591: StrReplaceFile corrupts undecodable bytes outside the edited region
### Summary

`StrReplaceFile` decodes the whole file with `errors="replace"`, edits the string, and writes the whole string back. Any byte in the file that isn't valid UTF-8 — anywhere, including far from the edit — is replaced by U+FFFD and written to disk as `EF BF BD`. The file's length and contents change outside the requested edit, permanently, and the approval diff can't show it because the diff is computed from the already-lossy string.

This is the same root cause as #2191 / #1952 (whole-file round-trip silently rewriting content the edit never touched), but for undecodable bytes rather than line endings.

### Where

`src/kimi_cli/tools/file/replace.py`

```python
# line 132
content = await p.read_text(errors="replace")
...
# line 170
await p.write_text(content, errors="replace")
```

### Reproduction

A file with a single invalid byte, unrelated to the edit:

```python
orig = b"alpha\nbeta \xff gamma\ndelta\n"        # 25 bytes
content = orig.decode("utf-8", errors="replace")
content = content.replace("alpha", "ALPHA")      # the requested edit
out = content.encode("utf-8")
```

```
before: b'beta \xff gamma'      25 bytes
after:  b'beta \xef\xbf\xbd gamma'   27 bytes
```

The `\xff` is gone and the file grew by two bytes, on an edit that only asked to touch `alpha`.

### One thing worth flagging before anyone fixes this

**Dropping `errors="replace"` from the write at line 170 does not fix it.** I checked — U+FFFD is a perfectly valid character to encode, so it still writes `EF BF BD`:

```
read=replace  write=replace  ->  b'beta \xef\xbf\xbd gamma'   25->27
read=replace  write=strict   ->  b'beta \xef\xbf\xbd gamma'   25->27   (identical)
```

The loss happens at the **read**. That said, line 170 does look inconsistent with the project's own rule in `tests_ai/test_encoding_error_handling.md` — *"Writing files and encoding Python strings to bytes do not require `errors="replace"`"* — and with the sibling tool, `write.py:158`, which writes without it. Worth tidying either way, just not as the fix.

`write.py` can't hit this bug: it reads with `errors="replace"` only to build a display diff, and writes `params.content` (fresh model-supplied text). `StrReplaceFile` is the only file tool that reads, edits, and writes back the same content.

### Possible directions

I didn't open a PR because the sensible fix is a design decision that's yours to make, and each option has a real cost:

1. **`surrogateescape` on both ends** — byte-exact round-trip (verified: 25 → 25 bytes, `\xff` preserved). But lone surrogates raise `UnicodeEncodeError` when the string is later encoded, which this codebase has already been bitten by once in #420. The edited content flows into `build_diff_blocks` and the approval display, so that risk is live here. It would also need `errors` widening in `kaos.path`, where it's typed `Literal["strict", "ignore", "replace"]`.

2. **Refuse the edit** — read strict, and return a `ToolError` when the file isn't valid UTF-8. Small, can't crash, and arguably right for a text-editing tool. Cost: it declines edits the tool currently performs (badly).

3. **Splice at the byte level** — apply the replacement to the original bytes and only rewrite the changed span. Most correct, most work, and would also fix the CRLF class in #2191 / #1953.

Happy to send a PR for whichever direction you'd prefer.

shoemoney:
## Related Issue

Resolve #2591

## Description

`StrReplaceFile` decodes the whole file with `errors="replace"`, applies the edit to the string, and writes the whole string back. Any byte that is not valid UTF-8, including bytes nowhere near the edit, comes back as U+FFFD and is written out as `EF BF BD`. The file changes outside the requested edit, and the approval diff cannot show it because the diff is built from the already-lossy string.

```
before: b'alpha\nbeta \xff gamma\ndelta\n'          25 bytes
after:  b'ALPHA\nbeta \xef\xbf\xbd gamma\ndelta\n'  27 bytes   # edit only asked to touch "alpha"
```

This PR detects the lossy decode and returns a `ToolError` instead of writing. The check runs before the approval request, so a corrupting edit is never offered for approval.

This is option 2 from #2591, refuse the edit. `surrogateescape` on both ends needs the `errors` type in `kaos.path` widened and is an exception to the convention in `tests_ai/test_encoding_error_handling.md`; byte-level splicing is a rewrite of how the tool applies edits. Refusing is small, cannot panic, and turns silent corruption into an error the user can see. The cost is that the tool now declines edits it currently performs by corrupting the file. If you would rather have option 1 or 3, say so and I will rewrite this.

Two design notes:

- U+FFFD in the decoded text is only a symptom, since the file may legitimately contain one. When U+FFFD is present, the raw bytes are re-read and strictly decoded to tell an original U+FFFD from a failed decode. A file without one still costs exactly one read. The strict decode is caught, not propagated, so it cannot panic on malformed UTF-8, and the read that produces the content the tool uses still passes `errors="replace"`.
- Detection cannot be `content.encode() != raw_bytes`. `kaos.local.readtext` opens without `newline=""`, so reads translate CRLF to LF, while `writetext` passes `newline=""`. That comparison would reject every CRLF file. `test_replace_allows_crlf_file` locks this down. The write normalizing those endings to LF is the separate bug in #2191 and is untouched here.

Not in this PR: `replace.py:170` writes with `errors="replace"`, which the project's own rule says is not needed on writes and which `write.py:158` does not do. After this guard it is a no-op, since the content is known to round-trip. Happy to drop it in a follow-up.

## Testing

Three tests added to `tests/tools/test_str_replace_file.py`:

| Test | Guards |
|---|---|
| `test_replace_refuses_file_with_undecodable_bytes` | The bug. Asserts the file is byte-for-byte unchanged. Fails on `main` (the edit succeeds and the file grows by two bytes). |
| `test_replace_allows_file_containing_real_replacement_character` | No false positive on a file that genuinely contains U+FFFD. |
| `test_replace_allows_crlf_file` | No false positive on CRLF files. |

Full suite: 2933 passed on this branch vs 2930 on `cbc15c07`, with the same 26 pre-existing failures (`tests/ui/test_usage.py`, `tests/utils/test_editor.py`) before and after. `ruff check`, `ruff format --check`, and `pyright` are clean.

## Checklist

- [x] I have updated the changelog (added the entry by hand under `## Unreleased`; `make gen-changelog` shells out to `kimi` itself, which I did not want to run against your repo).
- [ ] I have run `make gen-docs` to update the user documentation. No user-facing docs describe this behavior, so there was nothing to regenerate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Only write entries that are worth mentioning to users.
 ## Unreleased
 
 - Kosong: Stop sending an empty `anthropic-beta` header when no beta features are declared — adaptive thinking removes the interleaved-thinking beta, which previously left an empty header value that some backends reject
+- Tools: `StrReplaceFile` now refuses to edit a file that is not valid UTF-8 instead of silently corrupting it — the whole-file round trip replaced every undecodable byte with U+FFFD, anywhere in the file, including bytes far from the edit and invisible in the approval diff
 
 ## 1.49.0 (2026-07-16)
 
--- a/src/kimi_cli/tools/file/replace.py
+++ b/src/kimi_cli/tools/file/replace.py
@@ -131,6 +131,29 @@ class StrReplaceFile(CallableTool2[Params]):
             # Read the file content
             content = await p.read_text(errors="replace")
 
+            # This tool reads the whole file, edits the string, and writes the whole
+            # string back, so every undecodable byte in the file — including bytes
+            # nowhere near the edit — would come back as U+FFFD and be written out as
+            # EF BF BD. Refuse rather than silently rewrite bytes the edit never asked
+            # to touch. A U+FFFD present in the decoded text is only a symptom: it may
+            # equally be a real U+FFFD stored in the file, so confirm against the raw
+            # bytes before rejecting. The strict decode below is deliberate and is
+            # caught, not propagated, so it cannot panic on malformed UTF-8.
+            if "�" in content:
+                try:
+                    (await p.read_bytes()).decode("utf-8")
+                except UnicodeDecodeError as decode_error:
+                    return ToolError(
+                        message=(
+                            f"`{params.path}` is not valid UTF-8 "
+                            f"(byte 0x{decode_error.object[decode_error.start]:02x} at offset "
+                            f"{decode_error.start}). Editing it with StrReplaceFile would "
+                            "replace that byte, and every other undecodable byte in the file, "
+                            "with U+FFFD. No changes were made."
+                        ),
+                        brief="File is not valid UTF-8",
+                    )
+
             original_content = content
             edits = [params.edit] if isinstance(params.edit, Edit) else params.edit
 
--- a/src/kimi_cli/tools/file/replace.py
+++ b/src/kimi_cli/tools/file/replace.py
@@ -139,14 +139,15 @@ class StrReplaceFile(CallableTool2[Params]):
             # equally be a real U+FFFD stored in the file, so confirm against the raw
             # bytes before rejecting. The strict decode below is deliberate and is
             # caught, not propagated, so it cannot panic on malformed UTF-8.
-            if "�" in content:
+            if "\ufffd" in content:
+                raw_bytes = await p.read_bytes()
                 try:
-                    (await p.read_bytes()).decode("utf-8")
+                    raw_bytes.decode("utf-8")
                 except UnicodeDecodeError as decode_error:
                     return ToolError(
                         message=(
                             f"`{params.path}` is not valid UTF-8 "
-                            f"(byte 0x{decode_error.object[decode_error.start]:02x} at offset "
+                            f"(byte 0x{raw_bytes[decode_error.start]:02x} at offset "
                             f"{decode_error.start}). Editing it with StrReplaceFile would "
                             "replace that byte, and every other undecodable byte in the file, "
                             "with U+FFFD. No changes were made."
--- a/tests/tools/test_str_replace_file.py
+++ b/tests/tools/test_str_replace_file.py
@@ -246,3 +246,59 @@ async def test_replace_empty_strings(
     assert not result.is_error
     assert "successfully edited" in result.message
     assert await file_path.read_text() == "Hello !"
+
+
+async def test_replace_refuses_file_with_undecodable_bytes(
+    str_replace_file_tool: StrReplaceFile, temp_work_dir: KaosPath
+):
+    """A file that is not valid UTF-8 is left byte-for-byte alone."""
+    file_path = temp_work_dir / "invalid.txt"
+    # The undecodable byte is nowhere near the edit, on a line the edit never mentions.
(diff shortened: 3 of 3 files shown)


shoemoney:
Thanks — both Copilot comments checked against the source. Adopted one, and the other does not reproduce.

**Adopted: `"�"` over the literal glyph.** Correct, and the precedent is the repo's own — `tests/ui_and_conv/test_prompt_history.py:67` and `tests/ui_and_conv/test_prompt_placeholders.py:153` both write the sentinel as an escape. Mine were the only three literal glyphs in the diff. Fixed in `e318cc0`, along with binding `read_bytes()` to a name so it is obvious the strict decode and the error message read the same buffer.

**Not adopted: the `IndexError` on `decode_error.start`.** I could not construct one. `start` is always a valid index into `object` for a `UnicodeDecodeError` from the UTF-8 codec, including the truncated-sequence case the comment names:

```python
b"abc\xe2\x82".decode("utf-8")
# UnicodeDecodeError: ... invalid continuation byte / unexpected end of data
# e.start == 3, len(e.object) == 5  ->  e.object[3] is fine
```

I brute-forced it rather than argue from the docs: **47,883 distinct `UnicodeDecodeError`s** — every 1- and 2-byte sequence, a sampled 3-byte space over the interesting lead bytes, plus targeted truncations, overlongs, surrogates and out-of-range scalars (`\xed\xa0\x80`, `\xc0\x80`, `\xf5\x80\x80\x80`, `\xf4\x90\x80\x80`). **Zero cases where `0 <= start < len(object)` did not hold.** The decoder points `start` at a real byte by construction; there is no truncation shape that moves it past the end.

Happy to add a guard anyway if you would rather not depend on that, but it would be unreachable code and I would rather not add it silently.

shoemoney:
Worth recording here since it came up on #2591: this PR does **not** use a `content.encode() != raw` round-trip check, precisely because of the newline asymmetry @ayaangazali traced (`readtext` uses universal newlines, `writetext` passes `newline=""`). That check would reject every CRLF file.

The gate here is narrower: it only looks at the raw bytes when U+FFFD is already present in the decoded text, and then rejects solely on `bytes.decode("utf-8")` raising — so a CRLF file with no undecodable bytes never reaches the byte comparison at all, and a file containing a genuine U+FFFD is still editable (covered by `test_replace_allows_file_containing_real_replacement_character`). Line endings are untouched either way.

Still happy to rewrite as direction 1 or the byte-level splice if a maintainer prefers.

shoemoney:
This has been open since Aug 6 and no CI has ever run on the head commit. `commits/e318cc06/status` returns `total_count: 0` and the check-runs list is empty, so no workflow was triggered rather than one failing. Other outside PRs in this repo do have check runs on their head commits, so it looks specific to this PR.

The change makes `StrReplaceFile` refuse to edit files that are not valid UTF-8, with tests. Happy to rebase onto current main if that helps it pick up a run.

